Issue #335, opened 2014-10-10
midas open source product needs a logo


ultrasaurus:
People find innovation.gov confusing and expect something to be at that domain (as far as I know there are no plans for that)
![image](https://cloud.githubusercontent.com/assets/41906/4593363/65e1cc26-5085-11e4-90d6-50c44470f9d0.png)

I suggest with come up with a simple placehoder midas logo. It will be overridden in each deployment, but will let us quickly identify dev & test servers, without confusing new open source contributors.

ericadeahl:
Here's the logo on a white background
![midaslogo_transparent](https://cloud.githubusercontent.com/assets/5911280/5617421/93272514-94c3-11e4-9843-2000f6c2ecc3.png)

polastre:
:+1: 

at some point we should trace this in illustrator so we have high res versions.

ultrasaurus:
maybe @ericadeahl will add that to her queue -- then my request would be to make the # of stars match cabinet seats, which I think is 17 (15 + VP and President):

The Cabinet includes the Vice President and the heads of 15 executive departments — the Secretaries of Agriculture, Commerce, Defense, Education, Energy, Health and Human Services, Homeland Security, Housing and Urban Development, Interior, Labor, State, Transportation, Treasury, and Veterans Affairs, as well as the Attorney General. -- [whitehouse.gov](http://www.whitehouse.gov/administration/cabinet)

ericadeahl:
Here's the logo without the small text, at 72px height and 60px height. Let's see if there's an existing vector asset we can use - otherwise I'll redraw when I get a chance. 
![midas_transparent_3](https://cloud.githubusercontent.com/assets/5911280/5618112/5a1d0156-94ca-11e4-9baa-27bf13333864.png)
![midas_transparent_3_60px](https://cloud.githubusercontent.com/assets/5911280/5618114/5c65b7aa-94ca-11e4-88b9-1469a02d32a2.png)

ultrasaurus:
I sent email to @martymadrid to see if he has any original vector... tho it was quite a while ago now!

ultrasaurus:
Here's what it looks like in-place:

![image](https://cloud.githubusercontent.com/assets/41906/5618780/16191204-94d1-11e4-845a-5ffcda501671.png)

I made the area for the logo 60px high -- will submit a PR shortly.

polastre:
Hrm, the color palette of the logo doesn't really match the color palette of the site :confused:

[...]

PR #468, opened 2015-01-05
new logo for open source midas

ultrasaurus:
this is for issue #335 but we may want to leave that issue open for vector art


ultrasaurus:
@brandtheisey note: this increases the size of the space for the logo.  I attempted to make it be ok with smaller logo by using background-position to center vertically, but didn't test for that case.  Not sure if it affects CrowdWork

PR #477, opened 2015-01-08
Adding a Berksfile entry for chef-user.

DavidEBest:
Chef 12 introduced a breaking change by requiring metadata.rb
to contain a name attribute. The released user cookbook, 0.3.0,
does not contain this attribute, and therefore will not work
with the current version of chef. The name attribute is in
Github master, but the maintainers of the repo have not been
responsive about when/if the current version will be released.

This change switches the cookbook to use the Github master
version of the cookbook.


ultrasaurus:
Looks good to me!

PR #481, opened 2015-01-12
Specifying a waterline version to work around file upload bug.

### LINKED ISSUE #479: uploading photos fails on midas-dev
seems to work locally, might be a server config

DavidEBest:
Waterline version 0.10.15 broke the ability to save files specified
as a node Buffer to the database. Specifying version 0.10.13 will
allow uploads to work until they correct the problem.

Fixes #479 (albeit temporarily)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "description": "Innovation platform providing collaboration and crowdsourcing tools",
     "dependencies": {
         "sails": "0.10.5",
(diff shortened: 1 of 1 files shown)


DavidEBest:
Created an issue against waterline here: balderdashy/waterline#792

DavidEBest:
Waterline 0.10.16 fixes the bug, so this is unneeded.

PR #490, opened 2015-01-12
allow sails to alter schema in dev

### LINKED ISSUE #475: schema isn't automatically created anymore
Would be nice if ```make init``` could work without mucking with config files...

The [Sails deployment guide](http://sailsjs.org/#/documentation/concepts/Deployment?q=overview) says: "create the database on the server and then run your sails app with migrate:alter locally, but configured to use the production server as your db. This will automatically set things up."

One option is to check in our schema and load it as part of ```make init ```

ultrasaurus:
fixes #475

### DIFF
--- a/config/models.js
+++ b/config/models.js
@@ -15,12 +15,12 @@ module.exports.models = {
   *                                                                          *
   * See http://sailsjs.org/#/documentation/concepts/ORM/model-settings.html  *
   *                                                                          *
-  * Midas Note: Due to the implementation details of soft delete in our      * 
-  * fork of the postgres sails adapter                                       * 
+  * Midas Note: Due to the implementation details of soft delete in our      *
+  * fork of the postgres sails adapter                                       *
   * (https://github.com/18F/sails-postgresql), the migrate mode must be      *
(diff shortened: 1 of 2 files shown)


config/models.js
```diff
@@ -15,12 +15,12 @@ module.exports.models = {
   *                                                                          *
   * See http://sailsjs.org/#/documentation/concepts/ORM/model-settings.html  *
   *                                                                          *
-  * Midas Note: Due to the implementation details of soft delete in our      * 
-  * fork of the postgres sails adapter                                       * 
+  * Midas Note: Due to the implementation details of soft delete in our      *
```
DavidEBest: Could you update this comment (or just remove it) to reflect the new setting? You might also mention the sails treating it as 'safe' in production.

ultrasaurus:
oops. Will do!

ultrasaurus:
@DavidEBest wasn't exactly sure what to say -- wanna review and let me know what you think?

DavidEBest:
@ultrasaurus Looks good.

PR #491, opened 2015-01-12
make demo works now

ultrasaurus:
FYI: make demo failed in last release 0.0.27
don't have the fix yet, but this makes the error more readable


ultrasaurus:
latest commit fixed #488

DavidEBest:
Looks good to me.

Issue #486, opened 2015-01-12
As an admin, the number of task creators should be correct


ultrasaurus:
We don't have 19 task creators on midas.18f.us -- we only have about 7, maybe 9 at the most, but the admin dashboard shows 19:

![image](https://cloud.githubusercontent.com/assets/41906/5707132/1c4a5fb0-9a3b-11e4-87e5-c6279738f6c0.png)

dhcole:
Should this count include disabled users who have created tasks, or only currently enabled users? Unlike #485, it seems like for counting tasks, it's worth including disabled users, but wanted to confirm.

ultrasaurus:
I agree with should include disabled users

dhcole:
Could I get admin access to this instance of Midas?

PR #493, opened 2015-01-13
Filters user count on admin dashboard to include only active users

### LINKED ISSUE #485: As an admin, I should see total number of users (not including disabled users)
Quantity should be only *enabled* users

image

dhcole:
Closes #485

### DIFF
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -147,11 +147,11 @@ module.exports = {
               done(null);
             }
           });
-        }], function(err) { 
+        }], function(err) {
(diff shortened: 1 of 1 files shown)


dhcole:
@ultrasaurus This should cover the main task of filtering the "quantity" count in the admin dashboard. How about the "Who have created tasks:" count? Do we need to alter that as well?

dhcole:
@ultrasaurus nevermind. I realized, that's the point of #486

dhcole:
This should be good to go.

PR #494, opened 2015-01-14
Try different method for settings users.withTasks. Fixes #486

### LINKED ISSUE #486: As an admin, the number of task creators should be correct

dhcole:
Using an underscore (lodash) chain of operations to make it more clear how `users.withTasks` is counted.

Good to go.


ultrasaurus:
This looks good -- clearer code, no expected behavior change, right?

dhcole:
Right, no expected change. It *should* mean we won't see the issue captured in the screen shot in #486 again, though I'm not sure why that error initially occurred.

PR #501, opened 2015-01-15
Add wildcards for autocomplete query match pattern

dhcole:
With the .10 update for Sails, `LIKE` queries need to include wildcards. This adds them for the autocomplete endpoints.

Fixes 18F/midas-open-opportunities#57

@ultrasaurus ready for review

### DIFF
--- a/api/services/sources/model.js
+++ b/api/services/sources/model.js
@@ -28,7 +28,7 @@ module.exports = {
         // Set up the where clause
         var where = { where: { like: { } }, limit: field.limit || 10 };
         // Add the search term to find the field
-        where.where.like[field['name']] = term;
+        where.where.like[field['name']] = '%' + term + '%';
       }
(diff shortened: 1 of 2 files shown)


polastre:
Maybe we should include a test?

polastre:
Otherwise :+1:

ultrasaurus:
a test sounds like a great idea!

dhcole:
Added test.

ultrasaurus:
@dhcole test failed, but looks like one of those random Travis failures that happens sometimes.  I clicked the 'Restart Build' button

dhcole:
Weird. Looks good now

screen shot 2015-01-16 at 1 16 46 pm

ultrasaurus:
@dhcole yeah, that happens sometimes...  seems like npm just randomly fails now and then, which breaks ci

PR #503, opened 2015-01-16
since we run ‘make build’ on production

ultrasaurus:
we need additional packages
I added --production to the midas-cookbook npm install to speed up deploys
then it failed since we run 'make build' on production
this commit moves required libraries out of devDependencies


dhcole:
:+1:

PR #506, opened 2015-01-16
Set up global notification CC and BCC

### LINKED ISSUE #487: As an admin, I should get every notification
This should be configurable.  When people first deploy the app with a new set of users this is quite useful, but with any luck, we'll want to turn it off soon!

Don't forget to make a chef attribute for the configuration :)

dhcole:
If present, a list of addresses to CC or BCC will be added to all outgoing emails.

Use by setting the following in `config/local.js`:

```js
  // System-wide email notification settings
  // Set as a comma-separated list of email addresses
  notificationsCC  : '',
  notificationsBCC : '',
```

Closes #487

### DIFF
--- a/api/services/dispatch/sendSimpleEmail.js
+++ b/api/services/dispatch/sendSimpleEmail.js
@@ -26,6 +26,8 @@ function send (locals, html, text, cb) {
   {
     from: locals.from,
     to: locals.to,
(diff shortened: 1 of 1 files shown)


dhcole:
Requires 18F/midas-cookbook#9

PR #512, opened 2015-01-21
Change admin nav header markup to avoid logo and hover styling

dhcole:
@ultrasaurus 

screen shot 2015-01-21 at 12 31 38 pm


ultrasaurus:
this looks great!

Issue #496, opened 2015-01-15
As a user, my session should persist across server restarts


dhcole:
Currently, we're using the server's memory to store user sessions. This means if a server restarts, which happens during each code update, the sessions are erased, and all users need to log back in. This also means during development, developers need to log in each time they test new code.

A better way to go is to use a shared session store that persists across restarts. This would also open up the possibility of sharing sessions across multiple servers for a redundant / autoscaling cluster of web servers.

## Tasks

- [ ] find a shared session store we can use. Common options are redis, mongodb. Not sure what we can use from the AWS stack
- [ ] update [session configuration](https://github.com/18F/midas/blob/devel/config/session.js) to use the new store
- [ ] set a desired time for sessions to expire. Currently sessions also expire each time the browser is closed.

PR #514, opened 2015-01-22
[Not ready] Add persistent session store, increase session age

dhcole:
Increases the session length to 30 days and uses the main postgres database to store sessions. At some point, we'll want to use a separate database for sessions to increase performance, but this should be fine for now. It allows users to stay logged in across multiple web servers or server updates and reboots. 

It also makes development easier by persisting the developer's session over reboots.

Requires an initial setup task after `npm install`:

```bash
psql midas < node_modules/connect-pg-simple/table.sql
psql midas -c "GRANT ALL PRIVILEGES ON TABLE session TO midas"
```

Refs: https://github.com/18F/midas-cookbook/pull/11
Closes: #496

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -220,6 +220,12 @@ If you'd like to include a sample project, also run:
 
 In ```config/models.js```,  set ```migrate: 'safe'```
 
+#### Set up the session table in the database
+
+From the root of the midas directory:
+
+    psql midas < node_modules/connect-pg-simple/table.sql
+    psql midas -c "GRANT ALL PRIVILEGES ON TABLE session TO midas"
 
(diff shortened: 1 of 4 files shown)


dhcole:
@ultrasaurus For developers, I [added a note](https://github.com/18F/midas/blob/8f7b3fce7d6849a4a47914e74d97ceaeaa32aaad/INSTALL.md#set-up-the-session-table-in-the-database) in the install.md file about setting up the session table in the database. What's the best way to set these up automatically for our CI environments?

Related: 18F/midas-cookbook#1

dhcole:
I'm going to refactor this a bit after some config changes.

Issue #515, opened 2015-01-26
vagrant up doesn't set up database correctly (make init requires alter)


ultrasaurus:
We want to check in the current schema, so it's under version control, then make a tools/postgres/setup script that sets up database schema  (should fail on exisiting db)

ultrasaurus:
cookbook will also need to be updated

dhcole:
Closed by #518

PR #518, opened 2015-01-27
Add database schema management

dhcole:
We're at a point where we need to make some database changes, and we want to do so in a way that manages schema changes under version control and allows for automated deployment. This PR adds scripts and documentation for implementing a schema management process as well as managing schema changes going forward.

It adds the current database schema, migration scripts, and updates the make file and documentation for managing schema changes.

Enables: #514
Closes: #515

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,6 +61,26 @@ longer than 76 characters.
 The last part of the commit log should contain all "external
 references", such as which issues were fixed.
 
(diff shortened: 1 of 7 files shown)


ultrasaurus:
Tested this locally:
```
./tools/postgres/cleandb.sh 
make init
make demo
make test
make test-browser
sails lift
```
Got no errors, signed up with a new user, and successfully created an opportunity!

PR #520, opened 2015-01-27
Remove link to notification settings. Fixes #517

### LINKED ISSUE #517: share email should not refer to an URL that doesn't exist
When I look at an opportunity, then click 'share' in the upper-right, the bottom of the email has the text:

You can unsubscribe from these emails or change your preferences by managing your notification preferences at https://midas.18f.us/profile/settings

However, we don't have notification settings implemented, so we should remove that text

dhcole:


ultrasaurus:
Wonderful.

PR #523, opened 2015-01-27
add placeholder page

### LINKED ISSUE #521: Need a placeholder page
Let's set up a static html placeholder page for maintenance. We'll serve the page out of S3 and switch DNS to it from Route 53 during database updates.

dhcole:
Adds a simple placeholder page to be deployed to an S3 static host.

screen shot 2015-01-27 at 4 56 20 pm

Closes #521


ultrasaurus:
That's lovely. I think we prolly don't want to check in the open-opps logo into the midas repo.  Maybe check it into here:  https://github.com/18F/midas-open-opportunities  ?

dhcole:
Good call. Moving.

PR #524, opened 2015-01-27
Add modal static attribute. Add cancel button to new event modal.

### LINKED ISSUE #183: Clicking outside a model should not dismiss it (causing me to lose my work!)
When you click the gray area outside of a modal, it closes the modal and you use whatever work you've done.  Instead, the user should be warned that closing the modal will throw away their work and give them the opportunity to abort.

This is related to https://github.com/USStateDept/midas-crowdwork/issues/38.

Two possible options:
1. Add a warning as described above
2. @Chessen has proposed making it so that clicking the gray area doesn't do anything (you must explicitly click close, the X, or press the ESC key)

dhcole:
Adds a `backdrop: 'static'` attribute to modal dialogs to prevent closing them by clicking outside of the modal.

Closes #183


ultrasaurus:
looks good!

Issue #509, opened 2015-01-20
As user, logged in attempt count should reset after successful login


dhcole:
Looks like the user log in attempts doesn't reset when there is a successful login.

PR #525, opened 2015-01-28
Actually reset user's passwordAttempts

dhcole:
After validating a reset token, we need to reset the count of password attempts so the user may log in.

Fixes: #509

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -209,8 +209,18 @@ module.exports = {
             // destroy the existing token
             validToken.destroy(function (err) {
               if (err) { return res.send(400, { message: 'Error destroying existing token'}); }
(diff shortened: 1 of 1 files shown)


dhcole:
Depends on https://github.com/18F/sails-postgresql/pull/2

dhcole:
That error is just returned from the JSON response. The UI always shows this message for any error:

> An error occurred resetting your password. Please reload this page and try again.

screen shot 2015-01-28 at 2 01 32 pm

ultrasaurus:
Ok, I suppose that's fine then.  If reloading consistently fails, they will let us know.  I do wonder if we should be logging errors like this...

ultrasaurus:
opened separate issue for logging, will merge this one

Issue #532, opened 2015-01-30
midas-dev database needs make init


ultrasaurus:
We need to run ```make init``` manually since the automatic setup wasn't working when this DB was created:
![image](https://cloud.githubusercontent.com/assets/41906/5969274/7a90d362-a7d6-11e4-8032-b73384e96d40.png)

ultrasaurus:
verified fixed on midas-dev

PR #533, opened 2015-01-30
Fix make init

dhcole:
Two problems were occurring with `make init`:

1. Tests were passing even if responses from the API return error codes >= 400, which was hiding the second problem
2. At some point while handling the request, `req.body.id` is being added as `undefined`. When a model is created with an `undefined` `id`, it fails, because the `id` needs to be not `null`. If the `id` is not present, the database will automatically create one. I'm guessing this showing up now has something to do with the Sails v.10 upgrade.

This PR fixes both issues.

cc @ultrasaurus 

Refs #532

### DIFF
--- a/api/controllers/TagController.js
+++ b/api/controllers/TagController.js
@@ -47,10 +47,12 @@ module.exports = {
   add: function (req, res) {
     if (req.route.method != 'post') { return res.send(400, { message: 'Unsupported operation.' } ); }
     var tag = _.extend(req.body || {}, req.params);
-    // Trim whitespace
+    // Delete undefined tag id added to request
+    delete tag.id;
     if (_.isUndefined(tag.name) || (_.isUndefined(tag.type))) {
       return res.send(400, { message: 'Must specify tag name and type.' });
     }
+    // Trim whitespace
     tag.name = tag.name.trim();
     TagEntity.find({ where: { type: tag.type, name: tag.name }}, function (err, existingTags) {
       if (err) { return res.send(400, { message: 'Error looking up tag' }); }
--- a/api/controllers/TagController.js
+++ b/api/controllers/TagController.js
@@ -65,13 +65,13 @@ module.exports = {
         if (err) { return res.send(400, { message: 'Error creating tag' }); }
         return res.send(tag);
       });
-    });    
+    });
   },
 
   // Override default create to check parameters and
   // ensure duplicate tags are not created.
   create: function (req, res) {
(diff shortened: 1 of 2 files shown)


dhcole:
screen shot 2015-01-30 at 11 37 06 am

api/controllers/TagController.js
```diff
@@ -47,10 +47,12 @@ module.exports = {
...
     var tag = _.extend(req.body || {}, req.params);
-    // Trim whitespace
+    // Delete undefined tag id added to request
+    delete tag.id;
     if (_.isUndefined(tag.name) || (_.isUndefined(tag.type))) {
       return res.send(400, { message: 'Must specify tag name and type.' });
     }
+    // Trim whitespace
     tag.name = tag.name.trim();
     TagEntity.find({ where: { type: tag.type, name: tag.name }}, function (err, existingTags) {
```
polastre: @dhcole you were going to modify this to use `like` at some point right? (is that the subject of another issue?) either that, or you could simplify this using `findOne` instead of `find`.

test/demo/data/utils.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
     request.post({ url: url,
                    body: JSON.stringify(obj)
                  }, function(err, response, body) {
-      if (err) { return cb(err, null); }
+      if (err || response.statusCode >= 400) { return cb(err, null); }
```
polastre: I think this is okay for demo, but I wouldn't want this to make its way into the unit tests (for example).

api/controllers/TagController.js
dhcole: @polastre, this is just a quick fix since creating tagEntities through the /tag/add endpoint is currently failing. Switching to `like` is a separate issue...

test/demo/data/utils.js
dhcole: @polastre What's a better way to do it? The problem is the API is returning `{ message: "Error creating tag." }` with statusCode `400`, but the test is passing because it's getting a response. Shouldn't the tests check whether the server returns an error?

ultrasaurus:
Thanks for the quick fix @dhcole -- let's roll feedback from @polastre into future work!

test/demo/data/utils.js
polastre: Yes, but in unit tests we check for particular error response codes in the test case.  Which means you don't want `cb(err, null)` (which btw, `err` will be `undefined` so the messages will be useless).  Instead you want to pass the response back and let the test handle it.

PR #531, opened 2015-01-29
[Not ready] Draft Tasks Workflow

dhcole:
# Draft Tasks

This PR pulls together a bunch of different tickets related to setting up a workflow for draft tasks.

## Workflow

Admins set config options to determine the workflow for publishing new tasks. Tasks may either be created as `open` or `draft`. Once a new task is posted, [an email alert](https://github.com/18F/midas/issues/508) goes out to the task creator (admins CCd). The admin or task creator may then update their task to `open` when it's ready to be published to the site.

## Development tasks

- [x] allow tasks to be posted in draft state. 
- [x] set config option for default task state. 
- [x] show draft tasks in task list view to tasks' creators. Closes https://github.com/18F/midas/issues/467
- [ ] set config option for whether only admins can change task draft state. Closes https://github.com/18F/midas/issues/465
- [x] send notification to task creator. Closes https://github.com/18F/midas/issues/508
- [ ] add settings to midas-cookbook for deployment
- [ ] add email templates to open-opportunities

OO-specific notes are here: https://github.com/18F/midas-open-opportunities/issues/61

PR #534, opened 2015-01-30
Drafts Part 2

dhcole:
Continues work in #531

- [x] set config option for whether only admins can change task draft state.
- [ ] add settings to midas-cookbook for deployment
- [ ] add email templates to open-opportunities

### DIFF
--- a/api/controllers/MainController.js
+++ b/api/controllers/MainController.js
@@ -10,6 +10,7 @@ module.exports = {
     // include data variables for the view passed by the policies
     var data = {
       systemName: sails.config.systemName,
+      draftAdminOnly: sails.config.draftAdminOnly || false,
       alert: req.alert || null
     };
     // get version information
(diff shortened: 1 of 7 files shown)


dhcole:
Added new config setting `draftAdminOnly`. When `false`, the workflow is unchanged.

When `true`, only admin users may move a task out of draft state.

Task creators will see this:

screen shot 2015-01-30 at 2 11 24 pm

While admins will continue to see the usual button:

screen shot 2015-01-30 at 2 11 32 pm

With the option to move a task out of draft state:

screen shot 2015-01-30 at 2 11 45 pm

Once a task is no longer in draft, the task creator will be able to adjust the state:

screen shot 2015-01-30 at 2 12 01 pm

But not set it back to draft:

screen shot 2015-01-30 at 2 12 07 pm

Only the admin can do that.

brandtheisey:
:+1:

ultrasaurus:
Since my question isn't about new code, I'm went ahead and merged this pull request so we can start working with the new features.

Issue #309, opened 2014-10-06
after trying an e-mail address that already exists or is invalid, and then fixing it, all the error text and red colors should go away


louh:
1. Click "Sign in" and then "Sign up now" on the bottom right
2. Attempt to create a new account, but use an e-mail that's already there, or an email that does not look valid. Click "Register."
3. After you get the "The email address is not valid or already in use" error message, go back and edit the e-mail address so that is a new e-mail address, or is valid.
4. You should see a green checkmark on the right, but the red highlight and error message are still present. This is confusing!

![screenshot 2014-10-06 18 07 27](https://cloud.githubusercontent.com/assets/2553268/4534414/2ea474a8-4da5-11e4-9778-88c845ffe058.png)

dhcole:
Fixed by #539

PR #539, opened 2015-02-02
Issue 309 Register errors

rpylipow:
I took a stab at trying to help out with #309 (I came across it while browsing http://www.codeforamerica.org/geeks/civicissues). 

`make test` was passing without errors, but `make demo` had a few errors. I was unsure of differences?

COMMIT MESSAGE
- If user has entered bad information and
pressed 'submit', error messages appear.
This removes those messages as the user
fixes that bad information. Includes:

- if valid email is entered, removes
error messages

- if ALL password criteria are met, it removes
the error message.

- Created checkPasswordConfirm() method and
moved that error messaging below the confirmation.


dhcole:
Looks great. Thanks @rpylipow

rpylipow:
@dhcole Oh that makes sense. Thanks!

PR #541, opened 2015-02-03
Remove multiple email addresses from UI, fix admin editing permissions

### LINKED ISSUE #140: No way to change primary e-mail address
There doesn't appear to be a way to save any changes in the Account Information area of the user Profile. When I select a radio button to change the primary address, the Save Changes button isn't active. When I change something in one of the boxes above the Account Information area, the Save Changes button becomes active, but there is no change to which e-mail account is primary. Adding new accounts and deleting them works, but not changing the primary e-mail address

### LINKED ISSUE #449: When I edit a profile page, I should see the email address
As the owner of a page or an admin, I should be able to see the email address associated with the account.  All I see is:

image

### LINKED ISSUE #522: As an admin, editing a user's email address applies the address to the acting user's account (the admin)

dhcole:
This PR simplifies the UI for editing email addresses. The application only recognizes the most recent email address per user. Users can now edit this with a simple text box. Admins can edit any user's email address.

screen shot 2015-02-02 at 7 00 39 pm

Possible next step would be to refactor out the UserEmail model, but that may be a pretty big task and not required to fix these issues.

Closes #140 
Fixes #449 
Fixes #522


ultrasaurus:
@dhcole what do you think about adding a test that the changes actually affect the right user (whether made by the user or the admin) ?

dhcole:
Let's hold on this for a bit. I realized there's another problem, which is that while this changes the email address, it does not change the user name, which is also the email for local users. Need to think more about this, but it looks like removing the `UserEmail` model may be the way to go.

dhcole:
Ok this should be ready to go now.

@ultrasaurus I added several tests:

- change own email address
- verify user name also changes
- change user name as admin (which asserts the change and also that the user's email [does not equal the admin's](https://github.com/18F/midas/pull/541/files#diff-5546c6825656bc508fe4584061d133bdR149) who took the action)

PR #547, opened 2015-02-03
Set require for login.json to absolute path

dhcole:
@dlapiduz This replaces the relative path with an absolute path. Do you need this of all relative instances or just this one?

For instance, is [this](https://github.com/18F/midas/blob/devel/config/auth.js#L9) a problem?

```js
var authSettings   = require('./settings/auth.js');
```

How about [this](https://github.com/18F/midas/blob/devel/config/auth.js#L10)?

```js
var userUtils      = require('../api/services/utils/user.js');
```

### DIFF
--- a/config/local.ex.js
+++ b/config/local.ex.js
@@ -21,6 +21,10 @@
  * For more information, check out:
  * http://sailsjs.org/#documentation
  */
+
+// Set global entry directory
+global.appPath = require('path').dirname(require.main.filename);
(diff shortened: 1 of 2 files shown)


dlapiduz:
@dhcole I think that this PR solves the current problem I have... those other relative requires are not currently an issue...

dlapiduz:
so :+1: from me...

PR #557, opened 2015-02-04
Move DB migration to `make migrate`

dhcole:
`make build` no longer runs `./tools/postgres/init.sh`
Use `make migrate` instead.

cc @dlapiduz


dhcole:
Also includes a change to the `./tools/postgres/init.sh` script that sets the latest schema version after importing a new schema. This only applies to completely new databases.

PR #558, opened 2015-02-04
Reorganize configuration

### LINKED ISSUE #526: Move sails-postgres to NPM
Currently, we depend on 18F/sails-postgres as an external dependency. We should move this to NPM. Needs to link to the `softdelete` branch, which we can do [like this](https://docs.npmjs.com/files/package.json#github-urls).

Need to update the following:

- [ ] package.json
- [ ] install.md
- [ ] 18F/midas-cookbook

dhcole:
Accomplishes a bunch of reorganization tasks to make builds easier and deployments less error-prone.

- `18F/sails-postgresql` to `package.json` pinned to branch `softdelete`
- Remove all `.ex.js` files
- Remove all content from `local.js`, moved to specific config files. This frees up `local.js` to be used for only local settings that don't need to be checked into the repository
- Adjust build process and documentation to remove copying files around
- Adjust `.gitignore` to check in generic config files

Next steps would be to review the `midas-cookbook` process to make sure it works with this new configuration layout.

Closes #526


dhcole:
Closes #526

ultrasaurus:
:+1:

PR #562, opened 2015-02-06
Use postgres for session store

### LINKED ISSUE #496: As a user, my session should persist across server restarts

dhcole:
- Adjusts session cookie to 30 days
- Uses `sessions` table in postgres DB for active sessions
- Adds `connect-pg-simple` module for managing session through connect / express

Closes #496

### DIFF
--- a/config/session.js
+++ b/config/session.js
@@ -1,23 +1,3 @@
-var pgSession = require('connect-pg-simple'),
-    express = require('sails/node_modules/express'),
-    pg = require('sails-postgresql/node_modules/pg');
-
-// Build database config
-var environment = process.env.NODE_ENV || 'development',
-    configs = ['./connections', './env/' + environment, './local'],
-    config = {};
-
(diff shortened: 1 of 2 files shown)


ultrasaurus:
Looks good!

PR #563, opened 2015-02-06
Draft updates

### LINKED ISSUE #538: I should not be able to sign up for a draft task
Drafts are only visible for the task creator and the admin, so this doesn't cause any real problems, but it is confusing.

As an admin I see:
image

As a non-admin task creator I see:
image

I think we should just hide the "I'm in" button (or "Volunteer" in vanilla Midas) when in Draft mode.

dhcole:
- Hide volunteer button from task owners and when task is in draft. 
- Allow admins to see draft tasks associated with projects.

Fixes #538

(merge after #562 )


ultrasaurus:
:+1:

PR #566, opened 2015-02-09
Front-end tests

### LINKED ISSUE #330: need front end tests to verify '/tasks/1'
this recently broke (https://github.com/18F/midas/pull/327) -- we should make sure we have and automated test to make sure that our pages render.  This particular URL is probably the most important, after the home page!

dhcole:
Adds a test for `/task` and runs the browser tests on `make test`

screen shot 2015-02-09 at 5 37 52 pm

screen shot 2015-02-09 at 5 38 05 pm

Closed #330

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ start: server.PID
 
 server.PID:
 	sails lift & echo $$! > $@;
-	sleep 5
+	sleep 15
 
 stop: server.PID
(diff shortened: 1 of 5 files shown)


ultrasaurus:
looks good -- since this is only for tests, will merge without usual waiting period for feedback

PR #576, opened 2015-02-11
Update New Relic configuration

dhcole:
Sets up New Relic with simpler config. Just set the following environment variables to turn it on:

* `NEW_RELIC_LICENSE_KEY`: Your New Relic license key. This is a required
  setting with no default value.
* `NEW_RELIC_APP_NAME`: The name of this application, for reporting to New
  Relic's servers. This value can be also be a comma-delimited list of names.
  This is a required setting with no default value.

[More config options](https://github.com/newrelic/node-newrelic#configuring-the-module)


ultrasaurus:
I'd like to default to using the hostName for the NEW_RELIC_APP_NAME if not given.

Also, I'm a fan of how we use environment variables in ```auth.js```
```
      clientId    : process.env.LINKEDIN_CLIENT_ID  || 'CLIENT_ID',
```
then it is easy to see how to configure locally in development by just editing/overriding config files

PR #578, opened 2015-02-12
[WIP] Add Wercker file to build assets in CI

dlapiduz:
This is still Work In Progress but shows how it builds the assets and it pushes them to a branch of the config repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
         "chai": "*",
         "mocha-casperjs": "^0.5.3",
         "casper-chai": "^0.2.1",
-        "pg" : "~3.6.0"
+        "pg" : "~3.6.0",
+        "phantomjs": "^1.9.15"
     },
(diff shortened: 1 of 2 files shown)


ultrasaurus:
We no longer have the *.ex.js files
```
          for file in *.ex.js; do cp -n "$file" "${file/ex./}"; done
+          cd ../..
+          cd assets/js/backbone/config
+          cp -n login.ex.json login.json
```

dlapiduz:
@ultrasaurus removed that step per our chat with @dhcole

dhcole:
@dlapiduz Since we're not at a point of setting this up yet, I'm thinking we should close this. But we'll have it here in case we pick up integrating with Wercker.

Also, we not have `phantomjs` as a dev dependency instead of a main dependency so that we don't need to include it on the production servers.

PR #583, opened 2015-02-12
Ask for name before volunteering

### LINKED ISSUE #568: As a participant, if I try to sign up for a task without a name in my profile, I should be required to provide that info.
I'm thinking about a simple popup with a text box that asks the person for their name to avoid:

image

dhcole:
If a user doesn't have a name and clicks the Volunteer (or I'm in) button:

screen shot 2015-02-12 at 3 57 08 pm

Then continue with the sign up process:

screen shot 2015-02-12 at 3 57 26 pm

Updated name is saved to user's profile:

screen shot 2015-02-12 at 3 57 46 pm

Closes #568


dhcole:
@ultrasaurus ok, made those updates.

PR #591, opened 2015-02-17
Remove null attributes from user model

dhcole:
The DB returns null values for attributes, which sails then uses to overwrite existing values on update. This removes null values when the user model is loaded.

This should resolve https://github.com/18F/midas-open-opportunities/issues/60

### DIFF
--- a/assets/js/backbone/entities/profiles/profile_model.js
+++ b/assets/js/backbone/entities/profiles/profile_model.js
@@ -12,6 +12,12 @@ define([
 			this.initializeProfileGet();
 		},
 
+		parse: function(res, options) {
+			// Remove falsy values (db returns null instead of undefined)
+			_(res).each(function(v, k, o) { if (!v) delete o[k]; });
+			return res;
+		},
+
 		initializeProfileGet: function () {
 			var self = this;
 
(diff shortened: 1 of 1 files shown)


dhcole:
We don't really have a way to write a test for this yet, since it happens through the backbone application rather than the API and our browser tests do not currently support logging in -- they have no user accounts or database. That seems like a separate issue we'll need to address to improve testing.

dhcole:
Here's the SQL query we need to run to fix the missing user names:

```sql
UPDATE midas_user
SET username = email
FROM (SELECT "email", "userId" FROM useremail ORDER BY id DESC) AS ue
WHERE username is null AND midas_user.id = "userId";
```

@ultrasaurus how do you want to manage this? Should I just run it on the production DB? Save it in a shell script somewhere?

ultrasaurus:
@dhcole I think a shell script is a good idea, since we'll also want to add to release notes.  Maybe ```tools/postgres/set_username_to_email.sh``` -- ok with something less wordy if you have a better name -- would be good to have a comment that references this issue

ultrasaurus:
Also like the idea of this being a migration script, but wouldn't that require a schema version change which I don't think makes sense?

PR #593, opened 2015-02-17
Send default HTTP response for errors. Remove sails boilerplate error views

dhcole:
By default, Sails tries to send HTML templated error responses. Since our templates have variables that aren't made available on errors, we're seeing noise in the logs. This just sends simple HTML responses to errors, unless otherwise specified. It also removes the boilerplate Sails views for errors that we don't use.

Fixes https://github.com/18F/midas-open-opportunities/issues/72


ultrasaurus:
this will be helpful for diagnosing *real* errors -- thanks for taking care of this.

Issue #579, opened 2015-02-12
New text for Opportunity Assigned email


ultrasaurus:
In the following text will need to use "localized" versions of Task -> Opportunity, TaskOwner -> Opportunity Creator, etc.

```
To: Participant
CC: Task Creator
Subject Line: You’ve been selected for [task name here]! 

Hi [user’s name],

Great news: You’ve been selected to complete “[task name here]!”

Before you get started, please review the Opportunity details: http://openopps.digitalgov.gov/tasks/12

Have questions about what you need to do? Post them to the discussion — directly below the Opportunity description — for the quickest response. 

If you’d like more in-depth guidance, feel free to reach out to the Opportunity Creator (CC’d on this email) to set up a phone call or video chat.

Thanks again for joining us: We’re excited to have you on the team! 

Best,

The Open Opportunities team
```

dhcole:
@ultrasaurus Who should receive this email? There is not a concept of assigning a task to a particular person, just setting it as 'assigned'. Should all volunteers (anyone who hit I'm In) get this? Do task owners need a way  remove volunteers?

ultrasaurus:
Yes, everyone who is signed up should get this email.  Task owners and admins can remove participants (volunteers) by editing the task (at the bottom).

dhcole:
If there are multiple volunteers assigned, do you want a personalized email to go to each of them, or should it be one email with all volunteers in the "to" line?

If individual emails, should the owner be cc'd on all of them?

Currently, there isn't a mechanism for separate emails to multiple people on one notification, so if you want the first option, we'll need to build that to support this.

ultrasaurus:
see discussion in https://github.com/18F/midas/pull/601

ultrasaurus:
looks good!

PR #601, opened 2015-02-20
Task assigned email notification

dhcole:
This is mostly done but we need to decide who gets this notification and whether more development is needed around the assignment workflow. 

Discussion in #579

### DIFF
--- a/api/models/Task.js
+++ b/api/models/Task.js
@@ -41,20 +41,53 @@ module.exports = {
   },
 
   beforeUpdate: function(values, done) {
-    switch(values.state) {
-      case 'open':
-        values.publishedAt = new Date();
-        break;
-      case 'assigned':
-        values.assignedAt = new Date();
-        break;
-      case 'completed':
-        values.completedAt = new Date();
-        break;
-    }
-    done();
-  },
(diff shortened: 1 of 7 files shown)


dhcole:
Updated to send to all participants in the `to` line. If that's acceptable, then this is ready to go. If we need to send separate emails to each participant, this will require more work.

ultrasaurus:
I was going to suggest that approach.  However in the case of multiple recipients, not sure what value ```metadata.recipient.name``` will be.  Perhaps we should just omit Dear ______, in this case.

cc @LisaLNelson

dhcole:
I updated it to "Dear Participants"

https://github.com/dhcole/midas/commit/fb71651c13e71aa0e3d13d978ec8d17031e7d8b4#diff-d3e1aaff94abc154fa93fbc76f2cc871R1

ultrasaurus:
"Dear Participants" doesn't work for the singular case.  Just talked to @LisaLNelson and she agrees that just dropping "Dear ____" is good.  The email starts with "Great news..." which is a nice opener.

dhcole:
Ok, removed it fe1e164e281aec96e4950ba06ca80aabfd29bf47

ultrasaurus:
Thanks!

Issue #605, opened 2015-02-23
Browse Tasks returns all tasks not in draft state on 1st load


brandtheisey:
Browse tasks returns all tasks not in draft state on 1st load.

It should only return users draft tasks and open tasks or just open tasks.

dhcole:
Tried to get around this problem with a refactor in #656.

Supporting better pagination and filtering could be addressed in #863

dhcole:
This is by design now and optimized.

PR #606, opened 2015-02-23
[Not ready] Change default task browse state to open from not draft

brandtheisey:
The default browse state of not draft returns all task from the server and then is filtered down to only open tasks on the client side.

This change does this filtering on the server, increasing the responsiveness. It does not affect the search / filtering behavior once tasks are rendered.

Addresses #605

### DIFF
--- a/api/controllers/TaskController.js
+++ b/api/controllers/TaskController.js
@@ -28,13 +28,13 @@ module.exports = {
     // Only show drafts for current user
     if (user) {
       where = { or: [{
-        state: {'!': 'draft'}
+        state: 'open'
       }, {
         state: 'draft',
         userId: user.id
       }]};
     } else {
-      where.state = {'!': 'draft'};
+      where.state = 'open';
     }
 
     // run the common task find query


dhcole:
Would this prevent users from seeing tasks that aren't open or draft?

screen shot 2015-02-24 at 12 57 22 pm

brandtheisey:
It should. The filtered/search behavior and the initial browse load behavior are two reasonably separate processes.
In your SS the request should only return tasks in one of those three states.

ultrasaurus:
confused.  When I log in and click "Browse Opportunities" the default state is to show open tasks:
image

brandtheisey:
This causes the initial filtering to be done server side. In CrowdWork this means several hundred tasks and associated data are not sent before the filtering is done.

In Midas if you manually check /api/task, you should see that you get all non draft state tasks currently and then the client filters them out to only open. 

This removes an enormous amount of unneeded traffic on the initial page load.

ultrasaurus:
oh, right.  Yes, this is a good idea!

ultrasaurus:
To answer question from @dhcole, in https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/browse/views/browse_main_view.js
it appears that we do a new ajax request whenever filter options or search tags/text are changed.

dhcole:
Ok, I'm testing locally.

dhcole:
@brandtheisey curious what you saw locally when you changed a task state? For instance, when I move a task from open to assigned, and then check the assigned box on the task listing page, I don't see that task anymore. This makes sense to me because toggling the check boxes isn't requesting any new data from the server. Also, the way you've changed the query, all `/api/task` GET requests would be hardcoded to return only open or draft tasks.

Are you seeing something different locally?

I agree that filtering on the server side is a better way to go, but to do that, we'd need to pass a `state` parameter to the taskController to use in the where clause, and update the front-end code to issue new requests on filter change. Also, that state param would need to support multiple values for multiple check boxes selected in the filter UI.

Let me know if you want to talk through this more.

@ultrasaurus the `$.ajax` requests in that file are only to the search API for updating the search results and autocomplete suggestions, not the filtered tasks. That's still only loaded once on page load here: https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/browse/controllers/browse_list_controller.js#L84-L89

brandtheisey:
I thought that the filter recalled the search function with the new states making it separate from the call that is used on the base browse (first load). Let me test a few things to get my thoughts in order on this, but the intent was absolutely to make the first load show only open or user's draft tasks.

brandtheisey:
Nope you are right.

dhcole:
Cool just updated the title so we don't merge until it's updated.

dhcole:
@brandtheisey Looks like we can handle a lot of this filtering client side if we get back to the default Sails behavior: http://sailsjs.org/#!/documentation/reference/blueprint-api/Find.html


 Parameter      | Type         | Details
 -------------- | ------------ |:---------------------------------
 *              | ((string))   | To filter results based on a particular attribute, specify a query parameter with the same name as the attribute defined on your model. <br/> <br/> For instance, if our `Purchase` model has an **amount** attribute, we could send `GET /purchase?amount=99.99` to return a list of $99.99 purchases.
 where          | ((string))   | Instead of filtering based on a specific attribute, you may instead choose to provide a `where` parameter with a Waterline WHERE criteria object, _encoded as a JSON string_.  This allows you to take advantage of `contains`, `startsWith`, and other sub-attribute criteria modifiers for more powerful `find()` queries. <br/> <br/> e.g. `?where={"name":{"contains":"theodore"}}`
 limit          | ((number))   | The maximum number of records to send back (useful for pagination). Defaults to 30. <br/> <br/> e.g. `?limit=100`
 skip           | ((number))   | The number of records to skip (useful for pagination). <br/> <br/> e.g. `?skip=30`
 sort           | ((string))   | The sort order. By default, returned records are sorted by primary key value in ascending order. <br/> <br/> e.g. `?sort=lastName%20ASC`



Have you looked into what it would take to leverage that?

dhcole:
Here's what the blueprint route is doing behind the scenes. We could copy the important parts of this to our custom find function:

https://github.com/balderdashy/sails/blob/master/lib/hooks/blueprints/actions/find.js

brandtheisey:
Replaced by #624

PR #607, opened 2015-02-23
Stops app crash when tag data is incomplete

brandtheisey:
When browsing tasks, the application crashes if it attempts to access tags[i].tag when tag is undefined.

This appears to be related to relations defined in the tag table no longer existing or being corrupt. 

If the data is complete, this will have no effect. If the data is incomplete the browse will never complete, even if forever is working and catches the application crash.


ultrasaurus:
looks like a safe change that will make the system more resilient

PR #608, opened 2015-02-23
MyUSA Configuration

dhcole:
Addresses tasks related to MyUSA integration.

First round includes:

- If user has an existing account (based on email address) merge it with new MyUSA credentials
- Don't create multiple MyUSA accounts on subsequent log-ins.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ pids
 logs
 results
 
+.env
+
 npm-debug.log
 
 # Ignore modules, should be built using `node install`
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -29,8 +29,7 @@ function authenticate (req, res, strategy, json) {
       }
     });
   } else {
-    passport.authenticate(strategy, function (err, user, info)
-    {
+    passport.authenticate(strategy, function (err, user, info) {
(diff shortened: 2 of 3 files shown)


dhcole:
@ultrasaurus this + a change to the cookbook should have myusa auth working again. Can you let me know what you think of these workflow changes?

## Currently

Signing in with MyUSA created a new account every time. Sign in with MyUSA 5 times, you'd have 5 accounts. 

Signing in with MyUSA after you have already set up a local (username and password) account would create a new account.

## With this PR

Signing in with MyUSA only creates a new account once if no other account exists in the system.

Signing in with MyUSA will log you into an existing local account with the same username (email), bypassing the password authentication and linking the MyUSA credentials to your existing account.

---

If you're good with this workflow, than this should be good to merge.

ultrasaurus:
I like this workflow.

dhcole:
Ok, I made those updates.

ultrasaurus:
wonderful -- looking forward to trying this out!  Will wait to merge till ok by Travis, but should be fine

dhcole:
Great! https://github.com/18F/midas-cookbook/pull/27 also needs to be merged before deploying to the dev server.

Issue #101, opened 2014-01-13
No search results should still show something


polastre:
When you search for a tag but there's no projects or opportunities associated with it, it should still show projects.  For example, the screen should say "Nothing matches your search, however these projects might be relevant..." and then a series of project cards.

@azmiria or @sharonlo feel free to proposal IA/UX.

ultrasaurus:
proposal:  just add some hints here, in usability testing we found that sometimes people typed soemthijng like "how do I find an opportunity

-> @brandtheisey to propose some text

brandtheisey:
![screen shot 2015-02-23 at 5 57 26 pm](https://cloud.githubusercontent.com/assets/7915386/6339634/cb1597ce-bb85-11e4-86e2-26267655e7b9.png)

Is this the kind of idea we are going for?

ultrasaurus:
Yes, that's great!  We don't actually have a FAQ right now, but if you have one, just make the URL configurable and show that line if you have it.  Likewise, first line should only appear if Projects are enabled.

Thanks!

PR #609, opened 2015-02-24
Add text when no search results

brandtheisey:
Text displayed when there are no search or browse results to replace the current behavior of no feedback at all.

Uses uiconfig to decide to show Projects or FAQ lines.
Stores top text in translation file, so it can be customized per installation.
Uses uiconfig for systemEmail mailto for last line.
Is it's own template so a design can be implemented at a later date.

Addresses #101


ultrasaurus:
This is great!

PR #614, opened 2015-02-25
Timestamp patch

### LINKED ISSUE #613: 500 error when saving a task
reproducible on midas-dev.18f.us, here's what I see in the log when it happens:

```
    at Socket.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:427:10)
    at emitReadable (_stream_readable.js:423:5)
    at readableAddChunk (_stream_readable.js:166:9)
    at Socket.Readable.push (_stream_readable.js:128:10)
    at TCP.onread (net.js:529:21)
    at TCP.onread (/var/www/midas/node_modules/newrelic/node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31)

Details:  error: invalid input syntax for type timestamp with time zone: ""
```

dhcole:
Fixes #613 

Sails Waterline ORM treats `""` (empty string) as a string instead of `undefined`, so saving empty dates fail.

### DIFF
--- a/assets/js/backbone/apps/tasks/edit/views/task_edit_form_view.js
+++ b/assets/js/backbone/apps/tasks/edit/views/task_edit_form_view.js
@@ -211,9 +211,9 @@ define([
         var modelData = {
           title: this.$("#task-title").val(),
           description: this.$("#task-description").val(),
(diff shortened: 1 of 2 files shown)


ultrasaurus:
looks good!

PR #615, opened 2015-02-25
Locations

dhcole:
Support for dynamic locations.

- [x] suggest locations from geocoding API
- [x] replace location tags with location model associated to user profile
- [x] ui for adding new locations
- [x] add location ui to tasks
- [x] add location ui to profiles
- [x] add location ui to projects
- [x] CC BY 3.0 Attribution in UI where location data is shown
- [x] migration plan for existing locations (may not be needed given there are only a few locations in the production database)


dhcole:
Still a work in progress. Added location finder for profiles. 

screen shot 2015-02-27 at 5 29 20 pm


Still need to add this to tasks and projects.

dhcole:
Should be all set now. Just need to review the existing locations and figure out a good way to update them.

dhcole:
Multi-select locations:

multi-location

dhcole:
Here are the existing tags in production. 

screen shot 2015-03-03 at 10 54 03 am

I suggest the following:

- consolodate Washington DC tags, add geodata
- add geodata to Gaithersburg, standardize tag text
- remove open, worldwide, anywhere

@ultrasaurus Sound good to you? Do we need the anywhere tags? I figure if there's not a specific place, then just leave it blank.

ultrasaurus:
@dhcole for tasks, creators neeed to be able to specify "worldwide" or "anywhere" -- we just need one of those (I have a slight preference for latter, maybe @LisaLNelson wants to chime in on wording)

All the rest looks good!  I reviewed code, haven't run it, but love your animated GIFs :)

dhcole:
@ultrasaurus task creators have the option to specify worldwide when they make a task:

screen shot 2015-03-03 at 12 27 40 pm

Doing so saves the task without any location:

screen shot 2015-03-03 at 12 27 56 pm

The problem is some users are updating their tasks and specifying worldwide again:

screen shot 2015-03-03 at 12 28 57 pm

Or anywhere:

screen shot 2015-03-03 at 12 29 05 pm

Or open:

screen shot 2015-03-03 at 12 29 14 pm

But following the default behavior of leaving the location blank would accomplish the same thing without non-geographic locations in the database.

In fact, that's what the instructions under the location form already say:

screen shot 2015-03-03 at 12 57 53 pm

That's why I'm suggesting removing the "worldwide"-like locations and setting the location for those tasks as blank.

dhcole:
One option that might clear this up is to change the default value under "work locations" from "none" to "anywhere".

screen shot 2015-03-03 at 1 00 12 pm

ultrasaurus:
Sounds good. Also, let's be consistent and in the initial model, let's make it "anywhere" also.

dhcole:
Made text edits for "anywhere" and found a few "volunteers" to change to participants. Should be ready to go.

For those who have the code checked out locally, you'll need to run a `git submodule update` to update the select2 package.

ultrasaurus:
looks great -- excited about this new feature!

dhcole:
Per @ultrasaurus's testing, there's some usability improvements to be made around the profile location selector. I'll work on them today. Just flagging before we deploy this beyond dev.

PR #620, opened 2015-02-27
remove reference to local.ex.js - no longer used

### LINKED ISSUE #617: INSTALL.md shouldn't refer to .ex files that no longer exist
https://github.com/18F/midas/blob/devel/INSTALL.md

starting at "Copy the main settings files"

ultrasaurus:
fixes #617


dhcole:
Good catch

PR #621, opened 2015-02-27
script to fix db errors caused by pre-0.2.0 bug

ultrasaurus:
this script fixes the database errors that were caused by a bug fixed in 0.2.0
script needs to be run manually 
related to https://github.com/18F/midas/pull/591


dhcole:
:+1:

PR #624, opened 2015-03-03
[Not ready] Search / Filtering improvements

brandtheisey:
In #606, the default state for a first load of Browse Task was changed to Open instead of not draft to increase responsiveness of the initial page load. Testing of the PR#606 revealed a bug where  filtering behavior was inconsistent.

This PR causes a more consistent behavior with search / filtering changes.
--New AJAX requests are fired when searches or state filters are changed.
--if at least one state filter and one search term are used, results are only shown if both  searches produce results
-- if at least one search term and no filters are used, any results in the search set  will be used unbounded by state filtering.
-- if at least one state filter and no search terms are used, any task in the specified states will be used.
-- if no state filters and no search terms are used, then no results will be displayed.

Replaces #606 
Addresses #605

### DIFF
--- a/api/controllers/ProjectController.js
+++ b/api/controllers/ProjectController.js
@@ -72,7 +72,7 @@ module.exports = {
       processProjects( null, [] );
     }
     else {
-      Project.find({ where: { 'state': state }, sort: {'updatedAt': -1}}).exec( function (err, projects) {
+      Project.find({ where: { 'state': 'open' }, sort: {'updatedAt': -1}}).exec( function (err, projects) {
         if (err) return res.send(400, { message: i18n.t('projectAPI.errMsg.lookupPlural')});
         async.each(projects, util.addCounts, function (err) {
           return processProjects(err, projects);
--- a/api/controllers/SearchController.js
+++ b/api/controllers/SearchController.js
@@ -32,26 +32,28 @@ function search(target, req, res){
   // make the query well formed
   var q = req.body;
   q.freeText = q.freeText || [];
+  q.user = q.user || [];
 
   var stateSearch = function (search,cb) {
-    //this is a temp solution
-    // should be revisited once we are on sails 10
-    // we are waiting for support of contains in an or-pair fragment in waterline
-
-    var stateFrag= "";
 
     if ( search.length > 0 ) {
+
+      var states = [];
+
       _.each(search,function(term){
-        if ( stateFrag == "" ){
-          stateFrag = "'"+term+"'";
+        if ( term == 'draft' ){
+          states.push({ state: 'draft',userId: q.user });
         } else {
-          stateFrag = stateFrag+",'"+term+"'";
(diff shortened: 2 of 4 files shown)


dhcole:
Looks great. Just two things:

1. Can you look into removing uses of `Model.query()`? I know some of that predated the PR but we should be able to avoid it now.
2. It looks like drafts are coming through when open is selected. Can you look into how to filter drafts better?

screen shot 2015-03-03 at 4 22 35 pm

dhcole:
Once you've got those tackled, I'd like to run another round of testing against our production DB so we can see how much load time this shaves off. Nice work @brandtheisey

brandtheisey:
re: 2) when are you seeing that? in a default Browse tasks or during other state filtering /searching?

brandtheisey:
I cannot duplicate 2) in my environment at this time.

dhcole:
I see it when requesting `http://localhost:1337/api/task` which is what the default page load for `/tasks` gets. Once I mark as task as draft, they are returned in this request, even-though this should only return open tasks be default.

brandtheisey:
updated.

dhcole:
For posteriety, we're looking into the following performance issues, where the current code seems to perform about as fast as the new branch.

```
https://openopps-staging.18f.us/api/task (current `master` code)
2.468
2.578
2.459
````

```
http://localhost:1337/api/task (this branch)
2.640
2.128
2.052
```

Of course, we'd expect localhost to be faster naturally, but it's not _much_ faster. And this leads to a worse user experience because each time the filter is changed, users see a 2+ second delay.

It looks like the `tagentity` select queries are taking the bulk of the processing time.

Also, I'm seeing what looks like a race condition where if you turn off filters too fast (before the requests complete) you the list and filter get out of sync, and you end up with the wrong tasks in the list per filter setting.

brandtheisey:
I'm going to take a look at Waterline associations here and see if it makes a difference.

brandtheisey:
I think I found a place we can optimize before Associations. 

Note all this is using the current code and not my "optimizations".

In the service of listing 10 open tasks on my local machine there are, wait for it, **442 calls to taskUtils.tagAssemble**!!! 

The calls are roughly 50 / 50 to tasksId vs. userIds, many many many of which are duplicative For instance it calls tagAssemble with a where clause of userId: 3 a total of 140 times despite the fact that nothing by user 3 is currently open. User number 3 does have 140 tasks, but the system isn't displaying any of them.

dhcole:
Awesome. Would be great to do that as a separate PR so we can benchmark it independently.

brandtheisey:
To clarify I meant the existing optimizations in this PR, not new work.

dhcole:
@brandtheisey This is really good work and I think it makes sense to get it in. Probably should happen as a new PR after the tag reorg (#656). So let's close this one but keep the branch so we can reconcile with #656.

PR #625, opened 2015-03-03
[Needs review] Add optional modal to home page

brandtheisey:
Implements a CrowdWork required feature.

Adds a modal to the home page, only shows if modalHome.show == true in UIConfig && the logged in user has a key/value of showModalHome / true in usersetting.
Uses default text or values in the i18n translation files.
If user clicks the TOS at the end, the usersetting value is soft deleted and the modal won't show again on subsequent home page views.
If the user cancels the modal or doesn't accept the TOS, then they will see the modal on their next visit to the home page.


dhcole:
I think this looks great. Sorry for not reviewing it sooner. Do you want to take a stab at updating it for common.js requires (#662)? I can take a look if you get stuck with it.

brandtheisey:
Yeah, I'll do this today.

brandtheisey:
Replaced by #730

Issue #642, opened 2015-03-11
As a member who signed up before 0.2.3, I should be able to save my profile...


ultrasaurus:
and be re-directed to the view profile page...

ultrasaurus:
@dhcole investigating

PR #643, opened 2015-03-11
No location bug

dhcole:
Just a simple JS bug fix...


ultrasaurus:
code looks good, but can we do version change as separate?  I want to get to get the DAP fix in to the same release

ultrasaurus:
fixed https://github.com/18F/midas/issues/642

PR #647, opened 2015-03-17
[Not ready] Simplify project setup

dhcole:
Working on a few changes to the project to make it easier to set up and run.

- [x] Remove dependency on `make`. Use `npm run` instead
- [x] update documentation


brandtheisey:
I tested this locally on my mac and on my VM of server 2008 and everything appears to work so far. This should be able to run any of the make commands correct?

dhcole:
Thanks, that's great. It should run just about all of them, listed here:

https://github.com/18F/midas/pull/647/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L47

brandtheisey:
npm run init failed with this message on the server vm.

screen shot 2015-03-23 at 11 29 31 am

brandtheisey:
I guess I mis spoke, I tested npm run build first and that worked fine.

brandtheisey:
So far the issues I've found on Windows are:
1) in the test:* tasks, **mkdir -p .tmp/** fails because of the -p flag. Additionally  it is throwing a non-zero result if the directory already exists which causes the whole thing to show as an error even though I'm not sure it really is since there's no attempt to clean up the .tmp directory and it'll only not be there if you've never lifted sails.
2) in the test:* tasks it eventually calls **cp**
3) in the migrate tasks it fails if there is nothing to execute .sh scripts
4) n init & demo tasks, **export** & **mocha** fail because they don't exist.
5) import fails because **cp** & **rsync** 

Build and Build:dev both work correctly.

brandtheisey:
What will happen to grunt build if there is no .git installed?

dhcole:
This is now superseded by #662 but let's keep it open for sake of the windows discussion. That way, we'll know what's related to these changes and what's caused by the larger refactor.

dhcole:
@brandtheisey I just pushed 9203e679681307f7150a1b518febc68ce1598e50 to #662, which should address all of the Windows issues except `npm run import`. That requires more work because of `rsync`. Since you're not using that anyway (and most people probably won't / it may be refactored anyway) I'm included to punt on that for now.

Can you try the #662 branch on a Windows machine and let me know what you find?

Issue #649, opened 2015-03-18
Create a github page


ultrasaurus:
Target audience:  prospective contributors or people who are learning about Midas because they might want to install it themselves for use in their company or community

Draft Content:

### Midas is a “Kickstarter for people’s time.” 

Here’s how Midas works: You come up with a great idea that would help your workplace, but you and your coworkers don’t have the skills to execute it. Rather than bringing in a contractor, you can post your project idea to Midas, where other people within your organization can see it and then work on it, if they’d like. 

Midas is currently in active use [inside the US federal government](https://18f.gsa.gov/2014/07/16/midas-a-marketplace-for-innovation-in-government/) as the software that drives the [Open Opportunities](http://www.digitalgov.gov/join-digitalgov/open-opportunities-in-digitalgov/) program, which connects people from across and within government agencies to collaborate on projects. Open Opportunities validated the model using WordPress and a Lean Startup "concierge" approach, with people from over 50 government agencies, completing over 100 projects, ranging from 1 hours tasks to working part-time for 3-6 months.  Open Opportunities moved to Midas in October 2014, and is available for all federal employees [here](https://midas.18f.us/). Midas was originally developed by the Department of State and is being piloted there, specifically for foreign service officers.

Our Goal: to foster innovation across team boundaries by connecting individuals who identify challenges and then work collaboratively to implement solutions.

Midas can benefit all organizations, but it’s especially useful for those with distributed teams and whose employees have diverse skill sets.  Using Midas, you and your colleagues can work on projects together, get to know each other better, and develop skills you don’t use as often as you’d like.

How you can help:
* Use our [test server](https://midas-dev.18f.us/) and write up [issues](https://github.com/18F/midas/issues)
* Join our [Google Group](https://groups.google.com/forum/#!forum/midascrowd) and introduce yourself, please ask any general questions here -- how the product is supposed to work, usability issues, technical questions, whatever you want to know, we want to hear about.  If you are interested in using Midas for your organization, let us know.
* Read our [Contribution Guide](https://github.com/18F/midas/blob/devel/CONTRIBUTING.md) -- everything you do with us is a contribution to the public domain, there are also some tips for getting started
* If you are a developer, [install](https://github.com/18F/midas/blob/devel/INSTALL.md) the software for development
* Dive into our [help wanted issues](https://github.com/18F/midas/labels/help%20wanted) -- there are both dev and design opportunities there

ultrasaurus:
based on a first draft by @melodykramer

[...]

PR #651, opened 2015-03-18
Add static index that reads from README

dhcole:
Based on conversation in #649, this adds a simple home page for use as a github.io page that displays content from the readme.

To deploy, we keep the `gh-pages` branch up to date with `devel`. This could use some more styling, but should work for a first pass:

screen shot 2015-03-18 at 4 14 41 pm


ultrasaurus:
I love this!  Can you think of an easy way to markup the README header so it doesn't show up in index.html -- I'd probably omit the build status, and eventually make the logo part of a more functional header.  I could just move the build status lower, but its nice to be at the top of the README for the tech crowd.  (I'm sure I could find the elements with JS, but was wondering if marked.js had something we could just flip on.

dhcole:
Probably easiest way to hide them is just use CSS selecting on alt text. Just [pushed a change to do that](https://github.com/18F/midas/pull/651/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR22).

```css
img[alt="Build Status"],
img[alt="Dependency Status"] {
  display: none;
}
```

If we have to do a lot of hiding like this, we might want to think of something better, but if it's just a few things, this should do it.

PR #655, opened 2015-03-19
Update instructions in INSTALL.md.

GabeIsman:
These are steps that I found outdated and necessary, respectively, when going through the installation process.


ultrasaurus:
These look good to me :+1:

dhcole:
:+1: Thanks @GabeIsman!

PR #656, opened 2015-03-19
[Ready for review] Tag Associations

dhcole:
This pull request removes the `Tag` model and supporting code and instead uses [Sails Associations](https://github.com/balderdashy/waterline-docs/blob/master/associations.md) to link `User`, `Project`, and `Task` models to `TagEntity` models.

After this PR is merged, developers should run `npm run migrate` to convert existing tags to associations.

Moving forward, tags for models can be accessed with populate methods, like `User.findOne(1).populate('tags').exec(console.log)`.

Overall benefits for this change include faster queries for tags and loading for the `/tasks` page as well as a simpler API.

General changes include:

- [x] replace tag table with associations between `user`, `task`, and `project models with `tagEntity` models
- [x] migrate data from tag table to association join tables
- [x] update API code take advantage of the associations, populating attributes only when necessary
- [x] update the client application code to match API updates
- [x] update tests, `init`,  and `demo` commands


dhcole:
Finally have a sense of the scope. Here's what I think is left to make this shippable:

- [ ] Remove Tag model query in SearchController
- [ ] Make sure tags are set during user creation (for linkedin tags)
- [ ] Remove Tag model and controller
- [ ] Finish checking demo, init script, and tests

dhcole:
@ultrasaurus @brandtheisey This is ready for review! This update requires altering existing tag data to match the new schema, so please test on a database you don't mind messing up first and run `npm run migrate`.

I've been testing on a copy of our production database to test this at scale, but I'm interested in seeing whether you find any edge cases with different data.

brandtheisey:
:+1: 
I tested locally and it seems to work great. I added, searched and deleted tags and all the results were as expected. The only caveat being that I don't have a sort of live copy of the CW data to test it against locally so I was using the default midas test data.

But this looks great.

ultrasaurus:
tested locally, with latest commit, is works well for me :+1:

ultrasaurus:
@dhcole could you write up a release note and add to the top description of the pull request summarizing key changes?  for example: use of associations, that we dropped a db model, anything else we've covered a lot in standups that other devs should know when they upgrade

ultrasaurus:
@dhcole it's mostly in your checklist... but that might be a little too abbreviated for easy consumption

ultrasaurus:
This has been tested locally by @brandtheisey and me.  I think it is ready to merge as soon as @dhcole confirms that all the info that devs will need is captured in the description above.

PR #657, opened 2015-03-19
more welcoming text

ultrasaurus:
integrated text from #649


dhcole:
Looks great! Just one small change:

> ranging from 1 hours tasks  => ranging from 1 hour tasks 

Otherwise, all set.

dhcole:
Looks great! :+1:

PR #662, opened 2015-03-25
Client application refactor

dhcole:
Major refactor of the client application, aiming to make it easier to build and develop as well as for faster performance.

## High level goals
- Move from require.js to browserify
- Removes Make as a dependency in favor of npm scripts
- Removes git submodules
- Removes unused files
- Make application faster (using pre-built js package in development is much faster than 100 async client requests for require.js, and total bundle 

## Process
- Find each dependency that has an npm package and add it to `package.json`
- For packages not in npm, reference from github (if they have a package.json) or copy file to `assets/js/vendor/`
- Run [browserify-ftw](https://github.com/thlorenz/browserify-ftw) to replace AMD-style definitions with common.js `require` statements

    ```js
    // broswerify-ftw config
    module.exports = {
        quote           :  '\''         // '\'' or '"'
      , style           :  'var'        // 'var', 'comma', 'comma-first'
      , indent          :  2            // the tab size used in your project
      , directoryFilter :  null         // not supported yet
      , fileFilter      :  '.js'        // the extension of the file to upgrade
      , dryrun          :  false         // true|false if true no changes will be written to upgraded files
    };
    ```
- Global find and replace to change `require('text!filename')` to `require('filename.html')` and `require('json!filename')` to `require('filename.json')`
- use `browserify stringify` to read in text files
- Manually review dependencies and fix issues (some need to point to specific files in the package)
- Update `Gruntfile` to remove unused code and set up CSS build process to reference vendor css from npm packages
- Remove all git submodules and unused vendor packages
- Use a stub for jquery so packages reference globally included jquery (makes for better caching)
- Adjust `assets/js/backbone/app.js` to set up global dependencies
- update build scripts to run browserify on build

## Changes
I'll update the docs to reflect this, but here's the main things that change:

- no more Make (carried over from #647), instead run commands as follows:

    ```shell
    npm test         # runs tests. can also run test:api and test:browse separately
    npm start        # alias for `node app.js`
    npm run build    # builds the js, min.js, css, and min.css bundles
    npm run migrate  # runs the database migration script
    npm run init     # runs the initial data setup script
    npm run demo     # sets up demo data
    npm run import   # imports a directory of file for a new theme
    ```
- starting the application in `development` will automatically run the build task and watch for changes to the client application files
- starting the application in `production` will not try to build the client files, so they can be prebuilt and copied into place. To make sure the files are in place when running in `production`, run `npm run build` before `node app.js`
- new front-end packages should be installed with npm. adding files to `assets/js/vendor` is a last resort for libraries that aren't set up to work with npm
- assets are referenced with a query string that includes the git SHA if available, or the timestamp of the server run. Since this is a query string, we don't need to generate unique files and git is no longer a dependency for building the application
- performance: the bundled JS file is about 13% smaller than with require.js
  screen shot 2015-03-25 at 11 21 38 am
  screen shot 2015-03-25 at 11 21 24 am
- app set up should be as simple as `git clone ... && cd ... && npm install`

Closes #647


ultrasaurus:
Worked like a charm for me

```
git clone git@github.com:dhcole/midas.git dhcole-midas
cd dhcole-midas/
git checkout browserify
npm install
npm run test
```
all tests passed
```
npm start
```
seems fine so far.  


by the way, when you update the docs, can you put the vagrant section at the bottom?

brandtheisey:
All tasks pass on my local machine, :+1:

dhcole:
Great! I just pushed one more update with some potential Windows compatibility fixes for the build process, but I'll go ahead and merge this into `devel` and get it onto our dev server tomorrow morning. Thanks for testing!

PR #665, opened 2015-03-26
Fix the missing image problem

gboone:
http://greg.harmsboone.org/midas/


ultrasaurus:
ok, will merge and test it out!

Issue #660, opened 2015-03-23
as a sysadmin, I'd like to see timestamps in the application log


ultrasaurus:
I can't believe I never noticed this before, but must be sails default not to output timestamps

ultrasaurus:
looks like we can configure to use winston, which supports timestamps as an option.  SailsJS captainslog used to use winston -- possible we lost something with the recent Sails .10 upgrade

I'll give this a try

ultrasaurus:
FYI: We used to use forever in production, which adds timestamps to the logs.

I looked at using winston for logging (which allows configuration of timestamps among other things.), it seems like a good solution, but for some reason setting the configuration in config/log.js isn't having any effect.  

Just adding a note here, since I probably won't get back to this for a few days.  Below is the approach I tried.  Seems like a nice attribute of warden to be able to configure prettier console log and add timestamps to the file where you need it more...

```
var winston = require('winston');

var customLogger = new winston.Logger({
  transports: [
                new (winston.transports.Console)({
                    'level': 'info',
                    'colorize': true,
                    'timestamp': false,
                    'json': false
                }),
                new winston.transports.File({
                    'level': 'debug',
                    'colorize': false,
                    'timestamp': true,
                    'json': false,
                    'filename': './logs/test.log',
                    'maxsize': 5120000,
                    'maxFiles': 3
                })
            ]
});

module.exports = {

  colors: false,  // To get clean logs without prefixes or color codings
  custom: customLogger

};
```

brandtheisey:
I think there is a bug related to forever on Windows that might be the root of my issue.

brandtheisey:
I think this is the default config I would go with. In my opinion it's the simplest and easiest to understand . This and **npm install winston --save** is all that's needed I think. Of course we'll each have to change the filename to where our log paths go, but I think if we need more we can change this later.

```
var winston = require('winston');

winston.add(winston.transports.File, {
    filename: 'application.log', 
    level: 'verbose', 
    json: false, 
    maxsize: 5120000, 
    maxfiles: 3 
});

module.exports.log = {
  colors: false, // To get clean logs without prefixes or color codings
  custom: winston
};
```

PR #669, opened 2015-03-27
Add logging with winston

brandtheisey:
Basic logging with Winston. This should catch all current logging except the debug info thrown by i18n.

To use, you need to **npm install winston**

A couple of notes:
1) Winston uses different logging levels than sails (they are flipped), so the log_levels variable addresses that
2) Sails logging level (on line 68 of log.js) is the superset of available levels for other transports
3) The transport.Console is required to push info to the console and it will use the lesser of it's logging level and the sails logging level. Without it, you will see no logging information in the console.
4)Multiple transports of the same type can be created (if you wanted to write different levels to different files for instance) but they MUST have different **name** attributes or an error will be thrown.
5) The default log file needs to be given the appropriate full path when deployed.

Addresses #660

### DIFF
--- a/config/log.js
+++ b/config/log.js
@@ -1,5 +1,5 @@
 console.log('Loading... ', __filename);
-var winston = require('winston');
+//var winston = require('winston');
 /**
 * Logger configuration
 *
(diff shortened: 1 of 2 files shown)


dhcole:
@brandtheisey this is looking good. Two things:

1. Should we set the default log value to something less verbose?
2. Why have a file log by default? Can we keep only console logs by default and use a file case-by-case through the `local.js` file? If there's a good reason to do a file log, should it be in `.gitignore`?

brandtheisey:
@dhcole I believe this addresses your concerns as the file logging is now working correctly in local.js and the default is just console log with timestamps. In the comments of the log.js file are instructions on how to add the fulling logging to local.js.

dhcole:
@brandtheisey looks great! Any idea why so many old commits are in this PR now? Would be great to just have the commits related to this actual change so we can see a clean view of the history.

dhcole:
Works well. Thanks @brandtheisey

PR #671, opened 2015-03-27
Small setup improvements -- nvm to handle nodejs 0.10 dependency

sefk:
amend INSTALL: use nvm for installing node

Now that nodejs has moved on to 0.12 the default instructions (brew install nodejs) are no longer what you want. Midas depends on 0.10, for the time being at least. Instead of fragile instructions on versioning, recommend managing versions with nvm instead.

While in here, ignore vi temp files (*.swp).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ chef/cookbooks/*
 *~
 *#
 *.bak
+*.swp
 
 .DS_STORE
 .tmp
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -28,9 +28,14 @@ Start the postgres console acting on the midas database with: `psql midas`
     ALTER SCHEMA public OWNER TO midas;
     \q
 
-Then back to the command-line, install node.js:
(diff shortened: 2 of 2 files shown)


ultrasaurus:
:+1: looks good to me.

brandtheisey:
I like this, but we'll need to look for a windows option for this since we currently support both.

nvmw appears to have a git dependency and we are trying to go away from that.
nvm-windows looks self-contained enough from a surface reading.

@sefk do you happen to know about either of them?

But  :+1: , we don't need to hold up the PR while we figure out the other platform's issues.

sefk:
Thanks for the quick review and pull.

Sorry that I don't know anything about windows here -- indeed I'm new to node myself. So I don't have any wisdom to add on that front. Maybe someone tonight/tomorrow will know. We can watch for those dells/lenovos. :)

brandtheisey:
Was worth a shot just in case, I added a note to the Windows section. Thanks for the PR.

PR #674, opened 2015-03-27
Task like button could go negative

monkbroc:
Clicking like, then unlike would make the star disappear and numbers go negative.

like
unlike
negative like


brandtheisey:
:+1: I like this, but I'd suggest changing that href into a span and that will get rid of the weird underscore when you have focus (at least it does for me locally).

monkbroc:
Yeah, I saw the weird underscore. Using a span is a good idea.

monkbroc:
I changed the a tag into a button tag to make it consistent with the project like button. This makes the underscore go away but preserves nice things like tab order that span wouldn't have.

PR #676, opened 2015-03-28
Allow admin to delete events

brandtheisey:
Allows Admin to delete events.

At some point may expand to project owner  && event creator can delete.

This is a CrowdWork required feature.


dhcole:
Tested locally. Looks great. Thanks @brandtheisey

PR #679, opened 2015-03-28
Allow admin to delete comments

brandtheisey:
This allows admins to delete non-topic comments.

How to handle topics in general and topics with replies specifically will be addressed at a later date.

This is a CrowdWork requested feature.

### DIFF
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -293,7 +293,9 @@ module.exports = {
       newComment: function(event, done) {
         var activity = {
               type: 'newComment',
-              createdAt: event.createdAt
+              createdAt: event.createdAt,
+              comment: {},
+              userId: null
             },
             steps = [];
 
@@ -301,7 +303,8 @@ module.exports = {
         steps.push(function(done) {
           Comment.findOne({ id: event.callerId }).exec(function(err, result) {
             if (err) return done('Failed to find model' + err);
-            activity.comment = result;
+            activity.comment = result || {};
+            activity.comment.value = activity.comment.value || "";
             done();
           });
         });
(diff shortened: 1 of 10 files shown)


brandtheisey:
Removed topic restriction as we are moving to flat - comments in #693

brandtheisey:
If you deleted the only comment, the text encouraging you to start a discussion didn't appear, this fixes that.

ultrasaurus:
need to fix admin console that displays notifications (recent activity feed) -- can just say "comment deleted"

dhcole:
@brandtheisey take a look here: https://github.com/18F/midas/blob/devel/api/controllers/AdminController.js#L313

This throws an error when a comment has been deleted. I think we just need to check to see `if(comment)` and then return something generic like (deleted comment).

ultrasaurus:
I'm fine with just omitting the line with the deleted comment, with no annotation that the comment is deleted

dhcole:
Hey @brandtheisey any idea why there are a bunch of extra commits in this PR?

brandtheisey:
I can close this and resubmit a new PR, I'm not sure what happened.

dhcole:
@brandtheisey looking good. Found two things:

1. the "x" is a span and doesn't have the pointer cursor. Would this work as an `a` instead? Also, can you give it a width to make it a circle instead of oval?
2. deleting a comment shows default text even when there are other comments:

screen shot 2015-04-09 at 12 18 17 pm
screen shot 2015-04-09 at 12 18 23 pm

dhcole:
Just to recap from our call: @brandtheisey is working on the following:

1. Push updated X design
2. Stop default content from appearing when comments are present
3. Add note about a manual data migration for pre-existing threaded comments

brandtheisey:
@dhcole this has been tested to correctly handle pre-existing threaded comments and otherwise behaves as requested. There doesn't appear to be a need for a manual step.

brandtheisey:
Also closes #706

dhcole:
Looks great! Nice work @brandtheisey

PR #683, opened 2015-03-28
Click on opportunity state label changes the radio

monkbroc:
Simple UX fix. The labels next to the radio buttons should be clickable.

state


brandtheisey:
:+1:

PR #684, opened 2015-03-28
Don't migrate the file table automatically

### LINKED ISSUE #589: Sometimes, file upload data is being stored as JSON
Found this on the dev server. Profile photos were being stored in the file table as text, in what looks like json-encoded binary data. Needless to say this was causing the table size to balloon. Subsequent uploads went through fine. Not sure of the root cause yet. One way to just avoid the problem completely would be to not store files in the DB.

cc @brandtheisey Check to see if you have any records in your file table with large binary fields. Note: the size field reflected the size of the image before it was uploaded, which is when it blew up. So I was seeing records with a few kb in `size` but the actual binary data was several mb.

monkbroc:
This PR fixes #589 by never dropping and recreating the file table when the development server is started.

Something gets messed up in the extraction and re-insertion of the binary data during the automatic migration.

The real fix would be to dig deep in the database adapter and figure out what's going on, but this workaround is good enough since the file table is not likely to change often and thus need to be migrated in development.

### DIFF
--- a/api/models/File.js
+++ b/api/models/File.js
@@ -24,6 +24,12 @@ module.exports = {
     size: 'INTEGER',
     // Raw binary file data
     data: 'BINARY'
-  }
+  },
+
+  // Don't migrate this table automatically in development because the
+  // binary file data can get corrupted when it is extracted then
+  // reinserted into the database. Any changes to this table have to be
+  // migrated manually
+  migrate: 'safe'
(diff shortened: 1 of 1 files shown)


dhcole:
Smart workaround. Thanks!

monkbroc:
Welcome.

PR #687, opened 2015-03-29
Non-nested comment

monkbroc:
We propose to simplify comments by removing the hierarchy and making it easier for people to write new comments.

current comments

simpler comments


brandtheisey:
What about prepending every reply  with an @originalauthorsname and making that a href to the comment being replied to? That way to discussion would still be followable if multiple things were going on.

brandtheisey:
:+1: We like this a lot and are going to try out some variations of my comment above before this gets merged in.

dhcole:
This sparked a great discussion. @brandtheisey merged this into #693, which also includes the ability to block quote previous comments. This should balance the flat layout with the need to refer back to previous messages.

alan707:
:+1:

Issue #428, opened 2014-11-26
As a user, I should see icons that indicate a little better what they are for


shawnbot:
Some ideas @ultrasaurus and I batted around:

- [ ] Working Groups: [fa-comments-o](http://fortawesome.github.io/Font-Awesome/icon/comments-o/)
- [ ] Opportunities: [fa-rocket](http://fortawesome.github.io/Font-Awesome/icon/rocket/)
- [ ] Topics: [fa-tags](http://fortawesome.github.io/Font-Awesome/icon/tags/) (or [fa-tag](http://fortawesome.github.io/Font-Awesome/icon/tag/)) or [fa-newspaper-o](http://fortawesome.github.io/Font-Awesome/icon/newspaper-o/)
- [ ] Skills: [fa-cog](http://fortawesome.github.io/Font-Awesome/icon/cogs/)

PR #688, opened 2015-03-29
Adjust iconography to be more consistent.

phildini:
This is a potential fix for #428. This was a mostly find-and-replace job, as I wasn't sure if the ticket included adding icons where icons did not already exist. 

Suggestions or critiques welcome.


dhcole:
Looks great to me. @ultrasaurus this affects which tags are used to represent parts of the application. Want to take a look?

dhcole:
Thanks for the update @phildini !

ultrasaurus:
looks like what we asked for in the issue -- will merge.

Thanks @phildini

PR #690, opened 2015-03-31
The comment input should be full width

monkbroc:
While I realize that you have the hierachy of comments in there right now and the change to non-nested comments in #687 may not something you want to do right now, I still want to submit this UI fix for the width of the comment box.

comment field

I still think that a simpler comment system would help in the long run (I'm not sure I would know in which of the 3 boxes I should write a comment in the picture above).


alan707:
+1

dhcole:
Thanks @monkbroc. Merging this now for a quick fix. Will post an update on threaded comments on #687. Appreciate the fix.

Issue #667, opened 2015-03-27
As an admin, I would like to export CSV of user data


ultrasaurus:
Ideally this would be a link in the User Management panel of the admin section

ultrasaurus:
related to: https://github.com/18F/midas/issues/544

sefk:
See pull request #691 

![user_export](https://cloud.githubusercontent.com/assets/621565/6959998/4e16a8e2-d8d3-11e4-82f1-b2a9c0b4655b.png)

![user_excel](https://cloud.githubusercontent.com/assets/621565/6960026/8dd39d46-d8d3-11e4-94ee-8b0eb9ef6c3e.png)

Issue #668, opened 2015-03-27
As an admin, I would like to export CSV of task data


ultrasaurus:
Ideally this would be a link in the Task Management panel of the admin section
tasks.csv [id, name, description, created date, published date, assigned date, completed date, creator name/email/id, # signups]

sefk:
See pull request #691 

![task_export](https://cloud.githubusercontent.com/assets/621565/6960067/ee003170-d8d3-11e4-9462-c88be3b847a2.png)

![task_excel](https://cloud.githubusercontent.com/assets/621565/6960069/f2256f22-d8d3-11e4-8bf0-0a3b5e5505e9.png)

PR #691, opened 2015-04-02
Admin csv export for users and tasks

sefk:
Add a "download" button on the user and tasks admin pages to save a csv
format of those files. CSV file formatted for easy import into Excel by
handling null strings and formatting dates. Adds dependency on the
json2csv module to produce well-formatted csv (handles escape quoting,
etc.)

This fixes issues #667 and #668.

Some clumsiness on the task csv in how it counts volunteers to summarize
in the "signup" column. I couldn't get a groupby/count to work with the
waterline ORM, so had to resort to clunky sql and exclude that for test
cases, where the backend isn't sql-compliant.

### DIFF
--- a/assets/js/backbone/apps/admin/templates/admin_task_template.html
+++ b/assets/js/backbone/apps/admin/templates/admin_task_template.html
@@ -34,6 +34,15 @@
                 Completed: <%- completed.length %>
               </a>
             </li>
+
+            <!-- Export button -->
+            <div class="export-button pull-right tip" data-toggle="tooltip" data-placement="left"
+                 title="Download list of tasks as a comma-separated file ready for import into Excel">
+              <a href="/api/task/export">
+                <i class="fa fa-download"></i>
+              </a>
+            </div>
+
           </ul>
 
           <!-- Tab panes -->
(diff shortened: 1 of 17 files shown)


assets/js/backbone/apps/admin/templates/admin_task_template.html
```diff
@@ -41,85 +50,85 @@
...
+                  <a href="/tasks/<%- task.id %>"><%- task.title %></a>,
+                  <a href="/profile/<%- task.user.id %>"><%- task.user.name || task.user.username %></a>
+                  <% if (task.volunteers.length) { %>
+                  <p>Sign-ups:
+                    <% _(task.volunteers).forEach(function(vol) { %>
+                    <a href="/profile/<%- task.user.id %>"><%- vol.user.name || vol.user.username %></a>
+                    <% }); %>
+                  </p>
+                  <% } %>
+                </li>
```
sefk: Sorry about the reformats -- my IDE (webstorm) got a bit excited. Hope you don't mind.

dhcole:
@sefk This is really great! Thanks for tackling it. We'll review and get this into our next release.

dhcole:
I think this is great! We can revisit the SQL query later once we've moved to associations for volunteers. Thanks for tacking this @sefk !

dhcole:
And an extra :+1: for doing the tests too.

PR #693, opened 2015-04-02
Flat-comments w/ reply and markdown support

brandtheisey:
Alternative to #687 

Adds ability to quote/reply.
Adds marked support in the comment input.
Scrolls to comment box if you click reply and the comment box isn't in the viewport.
screen shot 2015-04-02 at 3 59 48 pm


brandtheisey:
screen shot 2015-04-06 at 11 18 13 am
Moved project owner list to left side so it's on the same side as it is on the opportunities page.

dhcole:
Looks great :+1:  

Thanks @brandtheisey  and @monkbroc !

Issue #670, opened 2015-03-27
As a task creator, I would like to see an introductory screen informing me of expectations


ultrasaurus:
on the /tasks page, when someone clicks [+ Opportunity], we would like an additional first step where the following text is presented

We’re glad you want to create an Opportunity! Before you continuing, please review the following Task Creator responsibilities.

As a Task Creator, I will:
* respond promptly and positively to task takers
* lay out clear and realistic deliverables
* be clear about how much time is needed to complete the task 
* define the skills needed 
* be prepared to onboard the task taker. If the task assigned is a simple one, take the time to explain it. If the task is big or long-term, be prepared to mentor the participant.

[  ]  I agree to fulfill these responsibilities as a task creator 
 (check box)

The state of the check box does not need to be stored in the database -- it would just be required to be checked before people continue to the next step

sefk:
See pull request #697. 

![responsib](https://cloud.githubusercontent.com/assets/621565/7009296/1e2202f2-dc4e-11e4-8aa5-460b4eda83b4.png)

Issue #695, opened 2015-04-06
New task wizard better button state handling


sefk:
The handling of the buttons in the New Task wizard is fragile, difficult to extend, and doesn't handle edge cases well. The code in modal_wizard.js calculates the button states (visible or not, disabled or not) in separate forward/backward functions, instead it would be better to leave the flipping logic in the forward/backward functions, but to have a single function to calculate and implement button state. 

- initial button state, don't rely on view template correctly hiding "create" button
- same behavior whether or not you get to something forward or backward
- don't enable "back" from first page

This is related to issue #670.

ultrasaurus:
:+1: this would be a helpful improvement to the system

brandtheisey:
:+1: I like this a lot. We have a lot of one off modal code and it would be great if it was more extensible.

sefk:
Fixed in PR #697

PR #697, opened 2015-04-06
Add responsibilities checkbox to "add task" wizard

sefk:
Fix issue #670: As a task creator, I would like to see an introductory screen informing me of expectations.

Involved a bit of hacking in the modal_wizard.js to handle the Save Draft button correctly. This works now, but could be done more generally, see issue #695.

responsib

### DIFF
--- a/assets/js/backbone/apps/tasks/new/templates/task_form_template.html
+++ b/assets/js/backbone/apps/tasks/new/templates/task_form_template.html
@@ -1,6 +1,6 @@
 <form id="task-form" class="form-inline" action="/api/task">
 
-  <section id="section-1" class="current nodraft">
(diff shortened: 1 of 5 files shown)


dhcole:
This looks great! Thanks again @sefk . Just one thing: we should change the references to the word "task" to use out i18n module (like `<span data-i18n="task">opportunity</span>`) since different implementations call them different things.

Do you want to take care of that or should I jump in?

sefk:
Thanks for catching that @dhcole, fixed.

Also did the modal button cleanup from bug #695, see commit 5363db5. I generally don't like putting too much into one PR like this, but it would have been a bit messy to sort out the dependencies.

dhcole:
Really great work and documentation in the commits. Thanks @sefk

PR #699, opened 2015-04-08
Change migrate to safe in the development env file as default

brandtheisey:
Since we use **npm run migrate** for db changes and since migrate alter doesn't play nice with our sails postgres soft deletes fork, migrate alter should never be used even in development.


dhcole:
:+1: `migrate: 'alter'` is nothin' but trouble.

Issue #631, opened 2015-03-05
As a member, editing my profile location, I expected it to let me search again..


ultrasaurus:
I searched on Washing  (expecting to find Washington, DC), but instead it found this (which is ok):

![image](https://cloud.githubusercontent.com/assets/41906/6513618/07ca5558-c331-11e4-9168-b60ce70de2a5.png)

Then when I tried to type "Washington" it shows "no matches found"
![image](https://cloud.githubusercontent.com/assets/41906/6513623/11ac5396-c331-11e4-807c-2ad5d53ba9c5.png)

I expected it to let me search again.

Workaround:  if I click away and click back into the editable field, then it works as expected

ultrasaurus:
I tested this locally.  It looks like the solution is simply to clear the text, which let's me re-type the query to do multiple searches.  Seems like a good approach @sefk

PR #705, opened 2015-04-11
location search in profile allow second searches

sefk:
Fix bug #631. When the search result doesn't have what you expect, enable a second search from where you were.


dhcole:
Works great for me locally. Nice job @sefk! Thanks for doing this.

ultrasaurus:
:+1:

Issue #672, opened 2015-03-27
As member, I would like to see where everyone is on a map


ultrasaurus:
We recently added locations with geo-locations, so now we should be able to display the community on a map.

From a development perspective, we could start by just creating a page at /map that displays a map of the world and where everyone is

From a design perspective, it would be great to have some mockups about how we might integrate this into the site

ultrasaurus:
I had a thought that we could let people pin themselves on the map as part of some prompts to fill out their profile at the beginning...  might be a fun intro to the site.

Added new related issue for this:  https://github.com/18F/midas/issues/700

sefk:
I'm still a bit hazy on the use case for this feature, @ultrasaurus. I can think of two places where someone would want to see location info.

1. Looking at a project, left pane shows avatars for "Requestor and Participants". That could be moved to the larger right pane and replaced with avatars on a map. 
- People without locations would float in the corner? That's an important case to handle well since initially nobody will have locations.
- Would want a visual indication of the requestor, maybe an avatar with a colored border?
- Would have to handle clustering well, as that's the typical case -- bunch of volunteers in Mexico City, bunch in Washington D.C., none anywhere else.

2. Looking at your own profile, see other people near you. On the profile page the right pane is already a bit tall. Clicking on a profile would take you to that person's page? If I'm making a new connection with someone 

In the meantime I'm going to do the table-stakes feature which is just a /map endpoint of all profiles and add location to the user model and profile page. That way we can see where *all* registered users are. But I'm not so sure that's useful in and of itself vs. these other use cases.

sefk:
I can use some design input, @dhcole and @ultrasaurus and @brandtheisey. I need to get all the users with their locations (technically, their location tag contents) to put on the map. There are a couple of ways to do this.

1. I don't see an endpoint to get all users, but I could add one. It could return a giant list of all the users similar to what you get from the /api/user call. Probably too much data to be useful, but most reusable. Portion of "user" response JSON below.

2. An all users endpoint, but with only location info to make it more sane. 

3. Add a specific "location" api controller with very specialized queries. What a really want is the list of locations, each with their own lat/long, with the list of who's at that location.

Or I'm open to other suggestions. Thanks.


```
{
"username": "sefklon@gmail.com",
"name": "Sef Kloninger",
"title": "",
"bio": "Test admin\ ",
"photoId": 1,
"photoUrl": null,
"isAdmin": true,
"disabled": false,
"passwordAttempts": 0,
"id": 6,
[...]

Issue #673, opened 2015-03-27
As a member, I would like a page where I can see other members of the community


ultrasaurus:
There's an old branch that had an initial implementation of this that might be worth checking out:  https://github.com/18F/midas/tree/feature/browse-people

The idea was that there would be a "people" tab and you could browse and search, and then people would be displayed with "cards" in a similar way to tasks.

Open to other design ideas on this.

arena:
I'm exploring some design ideas for this. Will post a link.

arena:
Exploration of this feature is here https://github.com/arena/midas-community-design

ultrasaurus:
@sefK is working on this, related to #672

dhcole:
Closed by #781

PR #716, opened 2015-04-16
D3 world map with users

sefk:
Partial fix to issue #672. Submitted here in case someone wants to take a look (pull branch, play with it). But probably needs some cleanup before asking for a thorough code review.

Text from commit below.

------------------
Add /map view that displays a world map with dots for all the users that have specified their own location. Currently colors and values are all hard-coded.

This is built off of mbostock's exceptional D3 library and borrows heavily from his examples. The world topo map is his 50m data file.

Add /api/user/all endpoint to dump out information for all users. Same access permissions as rest of the user endpoints.

This is a very rough initial version. 

- Efficiency of "all" endpoint needs to be improved
- Plot performance of lots of users is probably poor
- Plot performance of map itself on lower-end machines is probably not good enough, consider lower resolution maps
- Should add zoom, translate, etc.
- Needs spinner

Data flow through the map view is rough and should be refactored to take events into account.


ultrasaurus:
Would be worth looking into a smaller map data file -- I would guess the 50m supports a more detailed zoom than we need.

Thanks for posting this work-in-progress.  I will try to play with it later today!

dhcole:
> Would be worth looking into a smaller map data file -- I would guess the 50m supports a more detailed zoom than we need.

FWIW, I think this is a pretty standard geojson file for world maps. It should gzip down to around 120k, which isn't crazy for this kind of feature. I'd vote for getting this working well and then go back and optimize.

dhcole:
Actually, I just found about  a 110m file. Try this and see if it isn't too coarse: https://gist.githubusercontent.com/mbostock/4090846/raw/4efd24ad87179c1eeafd57cfb0978c8f45730036/world-110m.json

sefk:
It's interesting to compare the 50m and 110m maps. Screenshots of the two are attached, and I'm not sure which one I prefer.

The 110m, at first glance, is cleaner. Note the western coast of Canada and southern Alaska: instead of seeing a cluttered coastline, you just see the border. But then again, you are missing a ton of islands -- much of Indonesia, the south pacific, and the caribbean is missing!

world-50m.json - 727k
world-110m.json - 99k

50m
110m

dhcole:
Since you're not supporting zooming, I'd go with 110m.

sefk:
OK @dhcole that's what's in the code now. It's at a good stopping point. We can add a ton of future features later: zoom, pan, push position to navbar for linking, search on users, tooltips for user profiles, aggregate by city...

So: this is ready for code review. Thanks.

sefk:
Closing now that this will be merged into #673 .

Issue #719, opened 2015-04-16
when deleting an event, the text of the (x) button should not be selectable


ultrasaurus:
On Mac Chrome, I get a little I-beam text selection cursor when I roll over the (x) for an event (which I can do as an admin) -- I think this happens for most (all?) browsers and can make it "feel" like the button can't be clicked, but it actually works fine

ultrasaurus:
@brandtheisey: can you take a look at this? should be a quick fix, would be nice to get it into this release

brandtheisey:
Addressed in #720

PR #720, opened 2015-04-16
Change delete event span to anchor to match delete comment css/html

brandtheisey:
Change the css / html to match how delete Comments are marked.

Addresses #719


ultrasaurus:
haven't tested locally, but this looks like the right approach :+1:

dhcole:
Nice. Helps with screen readers too. Thanks @brandtheisey

Issue #640, opened 2015-03-11
As a member, I should not be able to save my profile with an empty name


ultrasaurus:
For historical reasons we prompt for a name:  https://github.com/18F/midas/pull/583

But we should also not let someone save their profile without a name.

PR #723, opened 2015-04-18
Issue #640: Prevent saving a profile with an empty name

csarsam:
Add client-side validation to profile editing that prevents members from saving with an empty name. Use utilities.validator for validation.

This should address issue #640.

As far as I can tell, there is no test coverage around this part of the application. Let me know if I'm mistaken and I'll add testing for this change.


ultrasaurus:
@csarsam welcome to the Midas project!  Thanks so much for this.  I haven't tested locally, but looks good on quick review.

dhcole:
Works great. Thanks @csarsam!

Issue #708, opened 2015-04-13
As a member, when I edit my profile and try to connect LinkedIn (when another account uses that LinkedIn), it does not actually connect the account


ultrasaurus:
1) Go tp my profile page
2) click "Edit Profile"
3) Click [+ Connect] next to LinkedIn
4) re-direct to OAuth page, type in name + password
5) re-direct to Profile edit page
still see [+ Connect] next to LinkedIn

If I save the profile, it does not appear to be connected

dhcole:
Started to look into this. Connect works for MyUSA but not LinkedIn. Moving out of the milestone so it's not blocking the release, but I'll assign this to myself to figure out.

dhcole:
@ultrasaurus is it possible you already have another account associated with your Linkedin profile? 

I can recreate what you're reporting when I have one account created by authenticating with Linkedin (which uses my gmail email address), then logging into my @gsa.gov account and trying to connect to my Linkedin account.

If I just connect an existing account to Linkedin with no other account associated with it, the accounts are connected.

The reason this is coming up seems like it's that we don't have our GSA accounts as our primary address for Linkedin, as opposed to MyUSA. When you have the same primary email, the credential would be added and you wouldn't be able to create two accounts with the same email address.

So what do you want to do about this? This is an edge-case but probably a common one. The easiest path would be to disallow signing in (and creating a new account) based on LinkedIn, but still allow for connecting Linkedin to an existing account.

Other than that... seems like we'd need to build some sort of feedback to the user that says "this Linkedin profile is already connected to another Midas account." But to make that a useful warning, we'd need to offer a way to merge accounts... which leads down a bloated path.

Can you think of other approaches that would be preferable?

ultrasaurus:
@dhcole that is exactly the issue.  It's a little weird since my LinkedIn account is disabled.

This won't happen on Open Opportunities since it never had LinkedIn authentication ON and that isn't planned, but we might still consider making it possible to untangle this with:
1) an error that says what is happening
2) make it so admin can delete an account, not just disable it

dhcole:
Cool. Those steps sound like a good approach.

How do you you want prioritize this?

dhcole:
@ultrasaurus I think this is covered by #729. Allowing admins to delete users can be a separate task.

PR #729, opened 2015-04-22
Show alert message when user tries to add an oauth account that's linked to another Midas account

dhcole:
This addresses #708 by showing the user an error to tell them their oauth account is already linked to another account. 

@ultrasaurus How's this work for you?

screen shot 2015-04-22 at 1 03 12 pm

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -19,6 +19,7 @@ function authenticate (req, res, strategy, json) {
       if (json) {
         res.send(userUtils.cleanUser(req.user, req.user.id));
       } else {
+        if (info && info.message) req.flash('message', info.message);
         res.redirect('/profile/edit');
       }
       return;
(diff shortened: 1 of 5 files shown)


ultrasaurus:
looks good to me

ultrasaurus:
wait.... "this profile" actually is confusing

ultrasaurus:
How about this:
"We're sorry, you can't connect that LinkedIn account since it is already connected to a different profile."

?

dhcole:
I was trying to avoid the extra work of figuring out what provider is throwing the error :) This could happen with LinkedIn or MyUSA, or any other oauth provider we may add in the future. I'll look into how we can set this value dynamically...

ultrasaurus:
alternate: "We're sorry, you can't connect an account that is already connected to a different profile."

dhcole:
Found it: `sails.config.auth.config.config[userAuth.provider].name`

screen shot 2015-04-22 at 2 34 33 pm
screen shot 2015-04-22 at 2 34 19 pm

ultrasaurus:
sweet :+1:

PR #730, opened 2015-04-22
Add optional modal to home page

brandtheisey:
This replaces #625 and incorporates all the suggested changes there.

(previous notes)
Implements a CrowdWork required feature.

Adds a modal to the home page, only shows if modalHome.show == true in UIConfig && the logged in user has a key/value of showModalHome / true in usersetting.
Uses default text or values in the i18n translation files.
If user clicks the TOS at the end, the usersetting value is soft deleted and the modal won't show again on subsequent home page views.
If the user cancels the modal or doesn't accept the TOS, then they will see the modal on their next visit to the home page.


brandtheisey:
Updated to use systemName and default to generic text.

dhcole:
Looks good. Do you mind adding the config instructions here? https://github.com/18F/midas/blob/devel/CONFIG.md

assets/js/backbone/apps/home/views/home_view.js
```diff
@@ -29,13 +41,116 @@ var HomeView = Backbone.View.extend({
       hostname: window.location.hostname,
       user: window.cache.currentUser || {},
     };
-
     this.$el.addClass('home');
-    compiledTemplate = _.template(HomeTemplate)(data);
+    compiledTemplate = _.template(HomeTemplate, data);
```
dhcole: Should this be in the format of  `_.template(templateHTML)(data)`? (http://underscorejs.org/#template)

brandtheisey:
Added instruction to the config.md document and made the change suggested by@dhcole. For some reason it appears to work both ways, but it's been changed for consistency and clarity.

dhcole:
:rocket: Nice work @brandtheisey, thanks for getting this synced up with the latest devel changes

Issue #717, opened 2015-04-16
As an admin, I would like to see agency and location in user.csv file


ultrasaurus:
current user.csv includes:
* user_id
* name
* username (email)
* title
* bio
* isAdmin
* disabled

related to:  https://github.com/18F/midas/pull/691

sefk:
I was going to take a look at this @ultrasaurus but I'd rather not until @dhcole's new tagging branch lands. Adding some dumb subqueries that will just have to be replaced with proper joins isn't smart.

dhcole:
Thanks @sefk. Tags work will be merged tomorrow. Also @csarsam already started on this task in #733 so I think we're all set here.

sefk:
I was so excited to try out your new branch @dhcole I went ahead and added these columns just now without noticing your comment or that @csarsam had done some work here too. I'll go ahead and submit my PR, since it's done, and you can decide which branch to pull.

Note highlighting in this screenshot is just to draw attention to new columns, that's not how the report actually exports. :)

![userexp-agency-location](https://cloud.githubusercontent.com/assets/621565/7357457/c4e90072-ece3-11e4-91ac-1ab540af3156.png)

ultrasaurus:
resolved with https://github.com/18F/midas/pull/733

PR #733, opened 2015-04-23
Add user location and agency columns to users.csv

csarsam:
This addresses issue #717 

---

*Change Summary:* 
The two fields 'location' and 'agency' have been added to the list of fields to export on the `User` model. Since the fields don't actually come from the user model itself, this seems a bit awkward, so I added a comment to note where they're populated. Let me know if there's a more appropriate way to handle it.

The `user/export` handler now retrieves all tags tied to the users being exported. The tags are mapped onto the user objects, with the tag type (ex: 'location) being used as the key and the tag name (ex: 'Tempe, AZ') being used as the value.

*Testing Changes:*
I updated the existing functional test on the `export` feature. Since I wanted to test that the columns were actually being populated, I do some setup in the test to make a new tag and give a user the tag as their location.

*How to Test*
* Server tests should pass
* As an admin, go to the Administration view and select 'Manage Users'. Select the export button and view the `users.csv` file that downloads. You should see 9 columns, including 'location' and 'agency', and the two new columns should be populated for users that have a location and/or an agency shown on their profile.


dhcole:
Thanks @csarsam! This looks great. We're testing #656 now, which changes the way tags work. Once that's in, we'll need to replace the [`tagUtils.assemble()`](https://github.com/18F/midas/pull/733/files#diff-2892a201d1e423fac8a79de57a600f3aR418). Tags will now be associated models, so you can just get them with something like `User.findOne(id).populate('tags').exec(function(err, userWithTags) { ... });`

I'll post back here after that's in so we can get this reconciled. Thanks again for tackling this.

clairesarsam-wf:
@dhcole Nice, that should eliminate all the awkward logic I have happening to get tags onto their associated users. Thanks for taking a look! When your changes go through I'll revise this PR.

csarsam:
@dhcole Nice, that should eliminate all the awkward logic I have happening to map tags onto their associated users. Thanks for taking a look, when #656 goes through I'll revise this PR.

dhcole:
@csarsam #656 is now merged. Just let me know if you run into any trouble with it while updating this work.

sefk:
So sorry to step on your toes here @csarsam! I didn't notice that you'd done this work in the old model before I blundered ahead and implemented my own pull request for this (#746) based on the new model. Take a look at what I did there -- was that how you would have done it?

csarsam:
@sefk No worries, these things happen! Probably should have assigned the issue to myself or something, not sure if there's a process I should be following for these things.

I took a look at your PR and it looks good - I like the change from `isAdmin` to `admin` in the exported sheet, and I agree that adding multiple api calls in the export test is not exactly clean. Personally since the export utilities were pretty generic before, I shied away from adding model/field specific handling methods to them and instead kept my tag onto user model mapping in the `export` handler in `UserController`.

I updated this PR to reflect my ideal solution - stole your `isAdmin` -> `admin` change & a comma cleanup in the admin test suite. Also removed the example location tag from the test data since I'm not using it currently.

ultrasaurus:
tested locally -- works great
code looks good too.  Love that @csarsam integrated ideas from @sefk, and that this was done test-first.   

Will merge -- thanks everyone :)

ultrasaurus:
This fixes https://github.com/18F/midas/issues/717

PR #738, opened 2015-04-24
Front-end security update

dhcole:
I did a full review of all client templates to make sure user-generated content is escaped. This should prevent cross-site scripting attacks (XSS).

- Reviewed all `<%= ...%>` and converted anything that didn't need to render HTML to `<%- ... %>` escaped tags
- Escaped user profile data before inserting it into the `backendUser` JS object
- Reviewed / adjusted all HTML template vars to make sure user-generated content gets escaped somewhere else (usually by `marked()` or using `_.escape(text)`
- Included a comment next to each unescaped template variable about why it's unescaped so we know that it's been reviewed and addressed upstream
- comment content is escaped on input instead of output to avoid conflicting with the autocomplete links

This is ready for review. Please keep an eye out for HTML code displayed in the UI that should be rendered (e.g. `<strong>Bold</strong>` vs **Bold**)

@ultrasaurus @brandtheisey @sefk


ultrasaurus:
@dhcole thanks for walking through all of the code in such detail!   Can you please add a few notes to ```CONTRIBUTING.md``` so new (and old) contributors remember to keep an eye out for these things in new commits and code reviews?

dhcole:
Great point. Added documentation here: https://github.com/18F/midas/commit/9ed4e5d65bbe48c3f9aa7433fee8dad36b2067c1

Issue #739, opened 2015-04-26
As a dev, when trying to initialize database - the following error is shown


kobekoto:
> midas@0.3.1 init /Users/kobeosei-akoto/sites/midas
> mocha test/test.upstart.js --development --recursive test/init

lifting sails: env=development
Loading...  /Users/kobeosei-akoto/sites/midas/config/application.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/assets.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/auth.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/auth.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/bootstrap.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/connections.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/controllers.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/csrf.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/email.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/globals.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/http.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/io.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/log.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/models.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/policies.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/routes.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/session.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/env/development.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/analytics.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/auth.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/emailTemplates.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/notifications.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/sources.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/settings/tags.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/sockets.js
Loading...  /Users/kobeosei-akoto/sites/midas/config/version.js
[...]

PR #740, opened 2015-04-26
Master

aifoS1:
Added a little more info regarding the running postgres and mentioned there may be an error when you run npm init.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,6 +19,15 @@ When Homebrew is done installing Postgres, follow the instructions at the end to
 Next, create the `midas` database:
 
     initdb /usr/local/var/postgresql
+
+Once you're done install you'll see two options:
+    Success. You can now start the database server using:
+
+    postgres -D /usr/local/var/postgresql
+    or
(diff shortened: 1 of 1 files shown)


dhcole:
Hi @aifoS1 thanks for these updates. The postgres section is great.

`npm run init` should work the first time you run it on a DB. You should only see an error if you try to run it more than once on the same DB. I actually just posted about this in #739 this morning.

Think we should update that line to be more specific that you should only run `npm run init` once per database, otherwise you'll see an error?

aifoS1:
Hi @dhcole, thanks glad I could help. Oh, interesting, I thought I ran it once, but maybe I did try to run it twice. Updating to be more specific regarding running it once can't hurt. Want me to update?

dhcole:
Yea, that would be great. Then we can get this merged in. I think a follow up task would be to have the `init` script return a more useful error, but documenting here should help alleviate the confusion.

aifoS1:
@dhcole great, updated!

dhcole:
Great, thanks again @aifoS1!

Issue #548, opened 2015-02-03
duplicate accounts created for what appears to be a single sign up


ultrasaurus:
With 0.0.30.1, Open Opportunities, we saw a user sign up with two accounts created -- the time stamps lead me to believe that this was a single user action that resulted in two records being created by the software:

![image](https://cloud.githubusercontent.com/assets/41906/6030197/b4116e12-abaa-11e4-9ff5-6af6a1d67296.png)

I crossed out the specific name, but it's same name, same email address.

ultrasaurus:
adding to the current milestone just because we're looking at a bunch of email issues and this might be addressed at the same time.

dhcole:
Couldn't reproduce, but we now have a test for this.

ultrasaurus:
This happened 4 times last week on Open Opportunities

dhcole:
I've reviewed the production database... I see no clues about why this is still happening. 

Also, I am unable to reproduce this locally, though we continue to see the issue in production.

The code around authentication could be simplified ID users only based on `username` (email) instead of emails in the `userEmail` table. 

To find the source of this, I'd need to do a full code review of the authentication process, which could take a full day or so. Alternatively, I should be able to enforce uniqueness on the user models, so if someone triggers a duplicate account, it would return an error. I've avoided this because it's not a great user experience, but it would stop the problem from occurring until we're ready to commit to reviewing / refactoring user auth.å

@ultrasaurus should I prioritize the full review now? In the mean time, I'll get the error set up.

dhcole:
@ultrasaurus this is in today's milestone, but based on our last conversation, we were planning to wait on a full refactor since this is an inconvenience for admins but should have no noticeable impact on users.

Should we move out of the milestone or do you see an actionable path to resolving this that we should do this week?

ultrasaurus:
cleared the milestone

ultrasaurus:
this continues to happen about once per week

dhcole:
Duplicate accounts may cause log-in problems because the email address is the same for both. If the disabled duplicate account has a higher ID number, auth attempts will query that account and block the user.

To fix, when disabling a duplicated account, we should also remove / change the email address to something arbitrary so that it won't conflict with the appropriate account during sign in.

dhcole:
Closed by #902

PR #744, opened 2015-04-27
[Not ready] Add unique constraint to User.username

dhcole:
This sets requires the `username` attribute of `user` models to be unique. If a new user model (account) is created with a `username` that matches an existing record in the database, and error will be passed on to the user:

> Unable to create new user. A record with that `username` already exists (`user's email address`)

This should put a hard stop to the issue that we keep seeing in #548. 

This PR also includes an additional catch to try to prevent the client from executing multiple account creation requests.

Requires running `npm run migrate` and should be merged after #656 

Finding and addressing the root cause would require a more thorough code review and likely refactor, as captured in https://github.com/18F/midas/issues/548#issuecomment-96748825

### DIFF
--- a/api/models/User.js
+++ b/api/models/User.js
@@ -12,7 +12,10 @@ module.exports = {
   tableName: 'midas_user',
   attributes: {
     // Login information
-    username: 'STRING',
+    username: {
+      type: 'STRING',
+      unique: true
+    },
 
     // Core attributes about a user
     name: 'STRING',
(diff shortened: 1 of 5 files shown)


ultrasaurus:
@dhcole won't we need to remove duplicate users from existing deployments before deploying this?

dhcole:
Holding on this per conversation with @ultrasaurus. There are two issues:

1. The constraint may block updating duplicate records, so we'd need to remove duplicates.
2. This stops duplicates but puts the error back on the user, which would be confusing since this problem usually occurs during one sign-up action. Better approach would be to log the error, but then try to sign in the user to the existing account and only fail if the password is wrong.

dhcole:
Going hold on this per @ultrasaurus's point that we should not make this the user's problem.

PR #746, opened 2015-04-27
add agency and location to user export report

sefk:
Add two columns derived from tags, relying on @dhcole's tag ORM refactor to do the heavy lifting. Fixes issue #717.

Note that while I was in here I changed the column name from "isAdmin" to just "admin," both to make it more clear, and to be consistent with the next column, "disabled." This might not be the kind of change that we'd want to do later, but I'm guessing that people aren't using this report much yet and it's early enough to make a change like that.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -343,7 +343,7 @@ module.exports = {
   },
 
   export: function (req, resp) {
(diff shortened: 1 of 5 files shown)


sefk:
Withdrawing this as it's redundant with PR #733

Issue #727, opened 2015-04-21
Make auth error handling consistent and robust


sefk:
Repro steps:

1. Log in as an admin user 
2. Select user name, "admin" in dropdown
3. Select user name, "logout" in dropdown

Symptom: ugly page with just "forbidden", screenshot below.

Should be: redirect to home page would be fine.

There are other places in the app where logging out doesn't give a great experience, for example when looking at the profile page. In that case you get an error ("not authorized") but at least are immediately prompted to log in with the login modal, so not as bad.

![screen shot 2015-04-21 at 10 17 55 am](https://cloud.githubusercontent.com/assets/621565/7258194/c8ce3352-e80f-11e4-9c5e-37ab25fdf14c.png)

ultrasaurus:
thanks for writing this up!

sefk:
I looked into this a bit and don't think there is a general way to handle, unfortunately. Access enforcement is handled by Sails in ```policies.js```, not in the controller itself as I thought. The fix would be going page by page and putting in redirects, more trouble than its worth. I propose closing this as wontfix. Thoughts?

ultrasaurus:
From reading various web articles, (like [this one](http://iliketomatoes.com/implement-passport-js-authentication-with-sails-js-0-10-2/) see step 8), it appears that we can do a re-direct from the policy itself.  However, our end-points are mostly APIs called from our single-page backbone app.  I wonder if we should do the "re-direct" client-side by navigating to the home page after log out and/or whenever we get a 403 from the server (like with requireUserId policy).

sefk:
That article is a good pointer @ultrasaurus. That points to a good approach, but the "requireUserId" policy is so pervasive, I don't believe the right thing to do is to change that policy everywhere.

Here's what I think we should do: add a new policy, "redirectWithoutUserId", that instead of returning 403, will redirect to the homepage, literally "/". We can use that policy for pages where staying on the current page makes no sense, like profile. We can use this for this initial case, and if there are others, we can switch incrementally. Doing it this way makes it easy to test. I'll code this up now and submit a PR.

sefk:
I was confused by backend / frontend. Indeed the backend already handles this correctly, returning a 403 when you're not logged in, so there's nothing more we want to do with sails or policy. The question is how to have the frontend respond to it. I now don't think that should be handled generally. For profile having it show as an error page and requiring login isn't so bad. The problem is with admin, once you log out there, there's no place for you to go.

Submitted PR #747 for the simpler fix.

[...]

PR #747, opened 2015-04-27
Better unauth handling @sefk

sefk:
This fixes issue #727. Currently logging out in the admin pages dumps you into an error page with nowhere to go. Now, redirect home. It's not perfect but I'm not sure what better place to send them.

Please review @dhcole or @ultrasaurus.


dhcole:
With #749, you can remove fc68bddca60b766c0f4085212807dae7434495f1

sefk:
More ambitious fix, handling errors globally now with ```$(document).ajaxError()```. The 397c473
 commit message:

    handle ajax errors globally with auth special case

    Part of fix for #727. Per discussion with @dhcole, instead of handling
    auth errors only in the admin section, this general handler helps
    for the whole app. Now whenever we get an auth error, fire up login.

    Add handling in the nav to prevent the login modal from opening multiple
    times.

    Instead of displaying a hardcoded error message, display the error we
    get back in the entity to the user. So those messages should be worded a
    bit better now: "You must login..." instead of "Access denied."

@dhcole: I still need your help with some event handling. The problem is with the dashboard admin page. Steps to repro:

1. log in as an admin
2. navigate to the admin pages, select the dashboard
3. open up chrome dev tools
4. select logout. You should see two 403 errors
5. the login dialog comes up, good. log in as a **non-admin user**
6. Now you can't log in as anyone anymore

Basically by logging in as a non-admin user on an admin page a *second time* you're causing yet another 403 error. The pending event to bring up the login dialog causes a problem.

I tried hacks like using ```_.defer()``` or ```_.debounce()``` but those weren't reliable, and wouldn't be robust anyway.

assets/js/backbone/apps/login/controllers/login_controller.js
```diff
@@ -59,11 +58,17 @@ Login = BaseController.extend({
...
         window.cache.userEvents.trigger("user:login:success", user);
-        self.cleanup();
       }).modal('hide');
+
+    });
+
+    // clean up no matter how the modal is closed
+    $('#login').bind('hidden.bs.modal', function () {
+      window.cache.userEvents.trigger("user:login:close");
+      self.cleanup();
```
sefk: originally these actions were only happening on successful login, i.e. not when the modal was dismissed by the "X". Factoring out into it's own listener fixed that.

dhcole:
:+1: looking good @sefk! I'll look into the multiple error issue this afternoon.

dhcole:
@sefk I ran through those steps this morning, but I'm not sure I'm seeing the same problem. Here's what I did. Note there is a js error for `user` not defined in the `$('.activity-block .activity-feed')` template.

1. Log in as admin
    screen shot 2015-05-04 at 10 43 15 am
2. Go to the admin dashboard
    screen shot 2015-05-04 at 10 43 36 am
3. Log out
    screen shot 2015-05-04 at 10 43 53 am
4. Log in as nonadmin (Get the log in dialog again, but logged in)
   screen shot 2015-05-04 at 10 43 59 am
5. (Click profile) Still logged in
screen shot 2015-05-04 at 10 44 13 am

sefk:
Now that I've merged off of current master which now includes @dhcole's fix (#766) this works fine for me in my dev env. I think it's ready.

dhcole:
Nice work @sefk! Tested, everything seems in order. Thanks for tackling this. :ship:

PR #749, opened 2015-04-29
Fix broken backend user data / missing admin link

### LINKED ISSUE #748: as an admin, if I reload the app on the admin page, I lose the "Adminstration" menu
Found locally, reproduced on https://midas-dev.18f.us version 0.3.2 (but not likely a new bug)

Steps:
1) Log in as an Admin
2) Choose "Adminstration" from top-right menu
3) You should be at: https://midas-dev.18f.us/admin/dashboard
4) Refresh the page

Results: "Adminstration" is missing from the menu 

Navigating elsewhere the problem remains, not fixed by a refresh:

image

dhcole:
Only escape string values. Move backend user code to controller instead of view.

Resolves #748

### DIFF
--- a/api/controllers/MainController.js
+++ b/api/controllers/MainController.js
@@ -12,7 +12,10 @@ module.exports = {
     var data = {
       systemName: sails.config.systemName,
       draftAdminOnly: sails.config.draftAdminOnly || false,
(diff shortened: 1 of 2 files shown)


sefk:
LGTM.

brandtheisey:
:+1:

PR #751, opened 2015-04-29
Use local jquery

dhcole:
Removes Google CDN jquery in favor of local minified jquery.

cc @konklone @msecret @brandtheisey


konklone:
Huge :+1:.

brandtheisey:
:+1:

PR #752, opened 2015-04-29
Enable CSRF tokens. Resolves #728

### LINKED ISSUE #728: Enable CSRF protection in backend and frontend
Midas is currently vulnerable to CSRF attacks. [Sails has good support][s] for csrf token issuing and checking, so that's good. That should be easy enough to turn on. On the backbone side it appears that the way is to [muddle with Backbone's sync event][b], maybe the destroy call too? That needs some playing with, of course.

  [s]: http://sailsjs.org/#!/documentation/concepts/Security/CSRF.html
  [b]: http://blog.softr.li/post/43146401263/finally-correctly-dealing-with-rails-csrf

dhcole:
CSRF tokens prevent a certain type of session attack. This turns on verification of a valid CSRF token for all non-`get` requests to the API, provides the token as an HTML `meta` tag, and adds the token as an HTTP header to all `$.ajax` requests.

More: http://sailsjs.org/#!/documentation/concepts/Security/CSRF.html

### DIFF
--- a/assets/js/backbone/app.js
+++ b/assets/js/backbone/app.js
@@ -4,6 +4,17 @@
  * running.
  */
 
+// Set CSRF header
+$.ajaxPrefilter(function(options, originalOptions, jqXHR) {
+  var token;
+  if (!options.crossDomain) {
+    token = $('meta[name="csrf-token"]').attr('content');
+    if (token) {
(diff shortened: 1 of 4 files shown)


dhcole:
Seeing no objections...

PR #755, opened 2015-04-29
Add populate method to Task controller findByProjectId method

### LINKED ISSUE #750: With .32 tags don't appear in the opportunity listing of the project page
It used to be like this:
tag_bug

Now it's like this:
screen shot 2015-04-29 at 11 34 03 am

It appears that the .32 tag changes might have caused this regression.

brandtheisey:
Added populate method so tags appear again in Project listing of associated tasks.

Fixes #750


dhcole:
:+1:

Issue #735, opened 2015-04-23
With flat comments, email text on comments needs to be adjusted


brandtheisey:
This:
```
<% if (metadata.parentComment) { %>
commented in your
<% } else { %>
started a discussion in your
<% } %>
```
should probably just be
```
commented in your
```
for everything.

PR #757, opened 2015-04-30
Minor email text changes

brandtheisey:
Minor text changes until we find an email text customization solution.
Addresses #735


ultrasaurus:
these look good

PR #742, opened 2015-04-27
Add email when volunteer deselected

brandtheisey:
This sends an email to a volunteer if they are deselected / deleted.

Uses text provided by @ultrasaurus .

Addresses #709

Issue #709, opened 2015-04-14
As a volunteer, I should be notified if I am de-selected as a volunteer


brandtheisey:
The act of an admin or owner deleting a volunteer should generate a notification indicating that the user hadn't been chosen for a the task they had volunteered for.

ultrasaurus:
Proposed text

To: Participant
cc: Opportunity Creator
Subject Line: An update on [Opportunity name here]

Hi [participant’s name],

Thanks again for your interest in [Opportunity name here]. We wanted to let you know that someone else was selected for this Opportunity, but plenty of others are still available! 

To see a selection of currently available Opportunities, please [click here](/tasks). 

Once again, thanks for your interest in this Opportunity. If you have any questions, please don’t hesitate to reach out.

Have a great day,

the SystemName team

PR #761, opened 2015-05-01
Add email when vol deselected

brandtheisey:
This sends an email to a volunteer if they are deselected / deleted.

Uses text provided by @ultrasaurus .

Repalces #742 

Address #709

### DIFF
--- a/api/models/Volunteer.js
+++ b/api/models/Volunteer.js
@@ -45,6 +45,33 @@ module.exports = {
     };
 
     noteUtils.notifier.notify(params, cb);
-  }
+  },
 
+  beforeDestroy: function (values, cb){
+   Volunteer.findOne({ id: values.where.id }).exec(function(err, volunteer) {
+      if (err) done(err);
+      var params = {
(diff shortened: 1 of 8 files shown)


api/models/Volunteer.js
```diff
@@ -45,6 +45,33 @@ module.exports = {
     };
 
     noteUtils.notifier.notify(params, cb);
-  }
+  },
 
+  beforeDestroy: function (values, cb){
+   Volunteer.findOne({ id: values.where.id }).exec(function(err, volunteer) {
+      if (err) done(err);
```
dhcole: @brandtheisey Should this be `cb(err)` instead of `done(err)`?

dhcole:
One comment about; otherwise looks good to me!

api/models/Volunteer.js
brandtheisey: You are correct, nice catch. Is fixed now.

dhcole:
:ship: thanks @brandtheisey

PR #762, opened 2015-05-01
CSS fix for truncated select2 inputs

dhcole:
Quick fix for a minor but annoying bug:

screen shot 2015-05-01 at 10 57 51 am
screen shot 2015-05-01 at 10 57 19 am

### DIFF
--- a/assets/styles/application.css
+++ b/assets/styles/application.css
@@ -226,6 +226,9 @@ input[type="submit"] {
   border-top-color: #c0c0c0;
   border-radius: 0 0 3px 3px;
 }
+.select2-container-multi .select2-choices .select2-search-field {
+  width: 100%;
+}
 
 .select2-container-multi .select2-choices .select2-search-choice {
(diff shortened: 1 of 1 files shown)


brandtheisey:
Oh this bugged the heck out of me :+1:

ultrasaurus:
yeah, I couldn't figure out why this was happening to me and not for other people when I noticed this a while back.  

Note:  open opps has its own [application.css](https://github.com/18F/midas-open-opportunities/blob/master/assets/styles/application.css) which appears to have the same bug

ultrasaurus:
@dhcole would you mind filing an issue, which we can keep open until its fixed in both places.

Regretting that I didn't create a separate css file for open opps with overrides... I had intended to refactor to use sass a while back.

dhcole:
Good catch. PR here: https://github.com/18F/midas-open-opportunities/pull/88

ultrasaurus:
Thanks!

PR #763, opened 2015-05-01
[Ready for review] Configurable fileStore for uploads

### LINKED ISSUE #589: Sometimes, file upload data is being stored as JSON

dhcole:
This PR sets up a `fileStore` service for storing and retrieving uploaded files. The replaces the existing practice of storing uploaded files as blobs in the database. The `fileStore` supports two services: `local` for the local filesystem and `s3` for remote hosting on AWS S3 — ideal for multi-server environments.

In addition to being a more scalable solution, this speeds up queries to the `file` table, which were the next slowest after `tagEntities`.

After pulling down this PR, run `npm install` to load the AWS-SDK as a new dependency.

## Developer notes

The developer API for `fileStore` is documented in `api/services/fileStore.js`, and it consists of a `fileStore.store()` method for saving files and a `fileStore.get()` method for retrieving files. Each service defines its own methods for each of these.

The `fileStore` is configured by `config/fileStore.js`, which is overridable in `local.js`. Configuration options are documented in that file. Specify the file hosting service with the `service` key. Each service has its own configuration object. For `local`, you may specify an optional `dirname`, the directory relative to the project root to store uploads. For `s3`, you must specify a `bucket` and you may specify an optional `prefix`. Both `dirname` and `prefix` should not use leading or trailing slashes.

## Data notes

This PR requires two changes: 1) a schema migration; 2) localizing the file data to the filesystem or S3 instead of the database. It does not delete data, though you may want to empty the `data` column of your `file` table to save space once you've confirmed this works as expected.

- run `npm run migrate` to update the schema
- run `node tools/postgres/localizeUploads.js` to localize file data

Fixes #589 for real.


dhcole:
@brandtheisey want to give this a shot to see how it works in your environment?

sefk:
## Install

Add to the dev notes for this that this needs an ```npm install``` before migration to pick up the new dependency on ```aws-sdk```.

## Localization

When I did the localization I got this console error:

```
{ [Error: ENOENT, open '/Users/sef/src/18F/midas/assets/uploads/1.png']
  errno: 34,
  code: 'ENOENT',
  path: '/Users/sef/src/18F/midas/assets/uploads/1.png' }
```

It's because the uploads directory didn't exist, so that file didn't get written. When I manually created and re-ran, then the error went away, but the file didn't get created. That file happens to be my avatar pic, I think, because now I get a 404 (broken link) on http://localhost:1337/api/file/get/1. Maybe localization isn't idempotent?

sefk:
Switching back to the devel branch fixed the 404. Good that you didn't remove the blob from the db, leaving that for a future step. So that's good.

dhcole:
@sefk @brandtheisey Thanks for catching the error if the directory is missing. I just pushed a commit that creates the directory if it's missing. Also the file localization now only adds the `fd` field to DB records if the file was successfully stored. This allows you to rerun the script in the event of an error.

You may need to manually empty the `fd` field for files that you want to localize since the previous script stored `fd` regardless of whether the file storage succeeded.

dhcole:
:ship: is leaving

PR #764, opened 2015-05-04
Allow task creators to clone tasks

csarsam:
**Change Summary**

*Server side changes:* This PR adds an endpoint to `TaskController`, `clone`. The endpoint accepts a task `id` and an optional `title` parameter (if no title is provided, the original title is used). It creates a new task instance, copying over the `projectId`, `title`, `description`, and `tags` from the original task. It then returns the id and title of the new task.

*Client side changes:* When viewing and editing a task, owners/admins now see an additional option:

Selection the option opens a confirmation modal that prompts for the title for the new task (defaulted to the original title with 'Clone' appended to the end). When the clone is submitted and succeeds, the user is navigated to the new task.

---

I'd like some feedback about the UI for task cloning - I'm not sure just sending the user to the new task is the right thing to do. Other ideas:
* Keep the modal open until the clone finishes, and ask the user if they'd like to navigate to the clone or stay on the current task.
* Show an alert stating that the clone has been made and that they can view it at any time in their 'Opportunities Created' list.

### DIFF
--- a/api/controllers/TaskController.js
+++ b/api/controllers/TaskController.js
@@ -58,6 +58,31 @@ module.exports = {
     });
   },
 
+  clone: function(req, res) {
+    Task.findOneById(req.body.taskId)
+    .populate('tags')
+    .exec(function(err, task) {
+      if (err) return res.send(err, 500);
+      // Create a new task, copying over the following from the original:
+      // projectId, title, description
+      // Don't copy over the task state or any dates
+      // TODO: Implement cloning tags as well
+      var data = {
+        title: req.body.title ? req.body.title : task.title,
+        description: task.description,
+        projectId: task.projectId
+      };
+
+      Task.create(data)
+      .exec(function(err, newTask) {
+        res.send({
+          title: newTask.title,
+          taskId: newTask.id
+        });
+      });
+    });
+  },
+
   export: function (req, resp) {
     Task.find().exec(function (err, tasks) {
       if (err) {
--- a/api/controllers/TaskController.js
+++ b/api/controllers/TaskController.js
@@ -58,24 +58,24 @@ module.exports = {
     });
   },
 
-  clone: function(req, res) {
+  copy: function(req, res) {
     Task.findOneById(req.body.taskId)
     .populate('tags')
     .exec(function(err, task) {
       if (err) return res.send(err, 500);
 
-      // Create a new task, copying over the following from the original:
+      // Create a new task draft, copying over the following from the original:
       // projectId, title, description, tags
-      // Don't copy over the task state
-      var taskCloneData = {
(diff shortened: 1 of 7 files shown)


brandtheisey:
I haven't had a chance to test this yet, but I like this alot.

One thought, does this need to have a policy controlling it? It seems like a good practice but I can also see it as being slightly redundant since anyone can make a task.

ultrasaurus:
This is exciting -- lots of people have asked for this!  I think it should be available to anyone who is logged in, just like [+ Opportunity]

Maybe we can come up with a better name than "Clone" ... ideas:
* New Opportunity from this one
* Save as...
* Duplicate Opportunity
* Copy Opportunity  

I think I like copy best. 

I would prepend COPY to the title, and also save as draft.  I think it should navigate to the task in edit mode.

We really, really need to add something to the draft state to distinguish it.

csarsam:
@brandtheisey I didn't think about that, we definitely should have policies in place. I'm going to switch to letting any logged in user use the feature, so I think just applying the same policies as for task creation should be sufficient?

@ultrasaurus Copy opportunity sounds great, pretty simple and clear. I'll make the titling/draft/view in edit mode changes as well.

Right now I'm showing the 'Copy' option from both show and edit mode for the task, I think removing it from the edit view will solve the issue of the user making edits and then being navigated away from the task before saving them.

One other potential issue - by allowing any logged in user to copy tasks, users who aren't owners of the project a task is in will be able to copy it. Should we just not copy over the projectId to the new task? That sounds annoying for users who ARE the project owner and want to keep making tasks in that project.

brandtheisey:
@csarsam I think you are right that anyone who can create a task should be able to clone / copy one. I think.

I'd copy over the project id if they are the owner and null it if not. But I also tend to overthink things.

dhcole:
> I think removing it from the edit view will solve the issue of the user making edits and then being navigated away from the task before saving them.

:+1: that sounds like the way to go

> Should we just not copy over the projectId to the new task?

I think that's the simplest solution. It's only one small step for a task creator to associate it with a project:

screen shot 2015-05-05 at 3 29 18 pm


Thanks for working on this @csarsam. This is a great feature to have!

csarsam:
@brandtheisey @dhcole I think I actually misunderstood how project permissions work - after looking at it more project non-owners seem to be allowed to add opportunities to a project, anyways.

The current PR status is that the project linking is not carried over to the copied task, I think it's probably the simplest option. If users find it inconvenient I can add it as an improvement later, but since right now even drafts show up in the list of tasks for a project I think it could get confusing.

dhcole:
Looks great! Thanks again @csarsam. :ship:

PR #766, opened 2015-05-04
If user is logged in and logs in with another account, log out first acc...

dhcole:
...ount before trying to log into the second. Fix duplicate oauth match ID.

Fixes a bug in https://github.com/18F/midas/pull/729 that prevented signing in with oauth accounts.
Addresses issue in #747 that occurs when logging into a new account while currently logged in.

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -13,6 +13,7 @@ var userUtils = require('../services/utils/user');
  * remote server.
  */
 function authenticate (req, res, strategy, json) {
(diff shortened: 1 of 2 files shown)


dhcole:
@sefk let me know how this works for you

sefk:
I tested this change on the branch for PR #747 and as advertised, it does fix the case I was seeing @dhcole. Thanks! I recommend pulling this one and then I'll rebase my branch and re-submit once it's merged. OK?

dhcole:
Thanks, and done.

Issue #769, opened 2015-05-06
process.env.PWD undefined on Windows


brandtheisey:
Code currently has two places where **process.env.PWD** is used to build a path. On windows this produces a file not found error because **process.env.PWD** is undefined.

The files are : config/application.js and config/settings/auth.js.

Suggestion is to change this to **process.cwd()** which works *nix,windows & mac os.

PR #770, opened 2015-05-06
Changed env.PWD to env.cwd()

brandtheisey:
--changes process.env.PWD to a more cross platform -> process.cwd()
Addresses #769


dhcole:
:fireworks:

PR #773, opened 2015-05-06
Location data fix

### LINKED ISSUE #768: saving location tags is no longer saving extra data (lat, long, source, etc.)
When user edits their profile and sets a loctions we need to save the extra data with the location tag -- lat, long, etc. To be stored in ```tagentity.data```.  I've confirmed that it is coming back from the autocomplete endpoint, just not getting carried along and set on the tag when its created.

In this screenshot you can see prior tags with the data column present, and the new ones without.

missing-tag-data-screenshot

dhcole:
Fixes  #768

- adds missing `tag.data` property when saving new tags to profiles, tasks, and projects
- adds a test for the tag creation API that verifies location suggestions and saving location data
- adds a script to update locations missing data.

If you have locations without location data, you can bulk update them by running `node tools/postgres/addLocationData.js` from the project root directory.

cc @sefk

### DIFF
--- a/assets/js/backbone/apps/browse/controllers/browse_list_controller.js
+++ b/assets/js/backbone/apps/browse/controllers/browse_list_controller.js
@@ -166,7 +166,8 @@ Browse.ListController = BaseController.extend({
       return _(tags).map(function(tag) {
         return (tag.id && tag.id !== tag.name) ? +tag.id : {
           name: tag.name,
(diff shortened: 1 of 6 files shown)


sefk:
I tested locally, both the migration script and profile page, and I see the data field carried along now. Thanks for the quick fix @dhcole !

Issue #776, opened 2015-05-08
In tag clouds / lists "Not Required" should not be rendered as a faux tag 


brandtheisey:
The third tag here, should be omitted in this list / cloud:

![screen shot 2015-05-08 at 10 37 02 am](https://cloud.githubusercontent.com/assets/7915386/7538550/cc8e229c-f56e-11e4-9289-a025dda3349d.png)

dhcole:
Fixed by #779

PR #779, opened 2015-05-12
Filter Required/Not Required from task list tag clouds

csarsam:
**Problem:** In the list of tasks shown on a project page, the `task-skills-required` tag was being rendered into the tag cloud. This tag is used to denote whether skills for a task are required or not, but isn't helpful to show in the task tag clound.

**Solution:** Filter out tags of the `task-skills-required` type before rendering the tag cloud.

This PR should take a task listing like this:

and turn it into this:


dhcole:
Looks great! Thanks @csarsam 

@brandtheisey, want to review this w/r to your issue in #776?

brandtheisey:
Looks good locally to me :+1:

PR #781, opened 2015-05-13
People view: show midas users on a map

sefk:
Solid start on issue #673. Show people in midas on a clickable map. Key features:

- Accessible via new /people page, currently not navigable while we decide how to roll out (soft launch)
- Display all users clustered by city, city dots drawn to not overlap
- Click on city to show list below map
- Tooltip on the map showing uses in city on hover. If lots of people there only show three sample names
- Map fits into container size given it and resizes cleanly
- Only authenticated users can see map or use new /api/user/all endpoint
- Map has its own controller / view with an eye toward embedding in other pages in the future (profile page?)
- More test data. If you run "npm run demo" it should be added to your existing database but YMMV. Myself, I nuked my test database and repopulated with my [midas-reset.sh](https://gist.github.com/sefk/b044a5dfeda3e0a9fdfd) script.

Future work

- Pan and zoom
- Show more detail in the list
- Filter users shown on map by filters (departments, search, etc).
- Consider other people visualizations (skill clustering, etc).

Ready for demo and code review tomorrow @ultrasaurus and @dhcole and @brandtheisey.

people-map-673


sefk:
@ultrasaurus saw bug where the tooltop was too far below and to the right on her display. Consistent for her but don't see on mine. Need to debug more.

sefk:
Feature: add avatar to list

test/demo/data/config.js
```diff
@@ -25,18 +25,91 @@ module.exports = {
       password: 'alanalanT1!',
       title: 'Deputy Secretary',
       agency: 'USDA',
-      location: 'DC',
+      location: 'Washington, DC',
       photo: 'assets/users/alan.jpg'
     },
+    'jon': {
+      name: 'Jon Snow',
```
ultrasaurus: I started to remove the game of thrones references.  We want to make the demo data sound like real-ish people who might work in the government with real-ish sounding titles, but not actually real people.

assets/js/backbone/apps/people/templates/people_list_template.html
```diff
@@ -0,0 +1,21 @@
+<table class="table table-condensed table-hover">
+
```
ultrasaurus: I prefer <ul><li> instead of tables, which is a bit more flexible in altering design for mobile, etc.

test/demo/data/config.js
```diff
@@ -25,18 +25,91 @@ module.exports = {
...
+      photo: 'assets/users/jaime.jpg'
+    },
+    'oberyn': {
+      name: 'Oberyn Martell',
+      username: 'oberyn@gameofthrones.com',
+      password: 'oberynoberynM1!',
+      title: 'Prince of Dorne and the Red Viper',
+      agency: 'GSA',
+      location: 'São Paulo, BR',
+      photo: 'assets/users/oberyn.jpg'
```
ultrasaurus: I usually get faces from:  http://uifaces.com/  -- we should file a separate bug, since they don't seem to be appearing.  I think they used to.

assets/js/backbone/apps/people/templates/people_list_template.html
```diff
@@ -0,0 +1,23 @@
...
+  <% _.each(people, function(person) { %>
+  <tr>
+    <td>
+      <a href="/profile/<%= person.get('id') %>">
+        <%= person.get('name') %>
+      </a>
+    </td>
+    <td><%= person.get('title') %></td>
+    <td><%= person.get('agency').name %></td>
+    <td><%= person.get('location').name %></td>
```
dhcole: @sefk looks like these `<%= ... %> should be HTML-escaped with `<%- ...%> instead, right? Applies to this and `people_map_tooltip.html`

assets/js/backbone/apps/people/templates/people_list_template.html
dhcole: I'm personally indifferent on table vs `li`s -- it should be whatever is most semantic for the presentation. In this case, @sefk has it set up as a table, so that seems fine. The default bootstrap tables has some media queries so it seems to reflow now.

dhcole:
@sefk Here's what it looks like against a recent snapshot of the production DB:

screen shot 2015-05-14 at 4 51 05 pm

This did expose an issue though, where we'll get a JS error from the underscore template for the user table if a user doesn't have an agency. Here's the call stack:

sefk:
That's nice to see, @dhcole, thanks for posting with the current prod data. It does show the problem that @ultrasaurus and I discussed yesterday, that tight clusterin in the Northeast will make it tough to select people that way. Unfortunately I don't see a good solution to that w/out implementing zoom and pan. If I can do that quickly then I will otherwise @ultrasaurus and I agreed we wouldn't want to block getting some user feedback on that.

If you can give me the call stack on the missing agency error I can make sure to repro/fix, thx.

assets/js/backbone/apps/people/templates/people_list_template.html
```diff
@@ -0,0 +1,23 @@
...
+
+  <% _.each(people, function(person) { %>
+  <tr>
+    <td>
+      <a href="/profile/<%= person.get('id') %>">
+        <%= person.get('name') %>
+      </a>
+    </td>
+    <td><%= person.get('title') %></td>
+    <td><%= person.get('agency').name %></td>
```
dhcole: @sefk seems like this is the source of a JS error when `agency` is undefined. Would probably apply also to the second line, though that's likely to not happen since you filter out users without locations.

sefk: of course, I'll just do ```person.get('agency').name || ''```

sefk: fixed, good catch @dhcole thx

sefk:
I've addressed all the bugs I know of and incorporated feedback. I believe this is read for another look @dhcole and @ultrasaurus.

dhcole:
Just ran through a local test again and it looks great! Nice work @sefk! 

@ultrasaurus good to merge?

ultrasaurus:
:+1:

sefk:
woot woot

PR #788, opened 2015-05-14
Add fd column to file table in current.sql

### LINKED ISSUE #787: file table missing "fd" column on initial DB creation
Error: "error: column file.fd does not exist"

Steps to repro
1. drop db, create new
2. npm run init
3. npm run demo
4. node tools/postgres/localizeUploads.js

The problem is ```tools/postgres/schema/current.sql``` doesn't have the new column. Migration 6 correctly creates it on a migrated DB, but not there on a new one, and we don't run migrations when creating a new DB from scratch. @dhcole can regen current.sh

dhcole:
Fixes #787

cc @sefk

### DIFF
--- a/tools/postgres/schema/current.sql
+++ b/tools/postgres/schema/current.sql
@@ -245,7 +245,8 @@ CREATE TABLE file (
     id integer NOT NULL,
     "createdAt" timestamp with time zone,
     "updatedAt" timestamp with time zone,
-    "deletedAt" timestamp with time zone
+    "deletedAt" timestamp with time zone,
+    fd character varying
 );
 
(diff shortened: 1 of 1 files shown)


sefk:
Diff looks fine and works in my dev env, :+1: 

I wonder how our tests passed without this column, btw. Maybe we just don't have test coverage around file operations?

dhcole:
Tests use a json-based database on disk so they don't interfere the data in your postgres DB. So we're not actually testing anything to do with the database. Not sure of the best way to fix that... probably at the integration layer, making sure that our staging db is up and accessible...

PR #794, opened 2015-05-18
Remove unsubscribe link from html email footer

brandtheisey:
Removes unsubscribe link from the HTML footer, because it points to a not yet implemented part of the user profile.


ultrasaurus:
:+1:

PR #795, opened 2015-05-18
Add nav bar alt text for user icon (508 compliance)

brandtheisey:
Adds alt text to the image for the user icon in the nav bar to make it 508 Compliant.

### DIFF
--- a/assets/js/backbone/apps/nav/templates/nav_template.html
+++ b/assets/js/backbone/apps/nav/templates/nav_template.html
@@ -33,7 +33,7 @@
         <li class="divider-vertical navbar-text"></li>
         <% if (user) { %>
           <li class="profile dropdown"><a href="/profile" id="navbar-profile" role="button" class="dropdown-toggle" data-toggle="dropdown">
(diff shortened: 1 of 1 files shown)


brandtheisey:
It does, however it was the Section 508 Program Manager who noticed and suggested this change. But I agree that your suggestion removes the issue.

brandtheisey:
Updated to say the user's name (or username) in the alt field.

This was suggested by a 508 Program Manager user of ours during the Mission Mexico Pilot.

dhcole:
@brandtheisey I got some guidance from our 508 guru, @nickbristow:

>  Images always need alt, but in this case, it should probably be
> ```alt=''```
> because it provides no additional information. or rather, the information is in text on the screen.
> 
> alt='Davide cole headshot' would work as well, but again, probably unnecessary

How do you feel about going with the empty alt tag so screen readers don't read "Photo of Dave Cole. Dave Cole."

brandtheisey:
Updated

dhcole:
:+1:

PR #802, opened 2015-05-19
Fix html new term suggestion in search w/escape

### LINKED ISSUE #799: Search displays html in autocomplete
When I search for a project, if some part of the text contains html, then the search display the html

Steps to reproduce:
1. Go to https://midas-dev.18f.us/tasks
2. In the search box, enter "3D"
3. Note the result returns text with html

brandtheisey:
- adds escape to new (unmatched) search terms in the select2 drop down
- renders html correctly for new (unmatched) terms in the select2 drop down

This crept in when escapes were added.

Closes #799 

Thanks for the report @nhu313


brandtheisey:
Adjusted to actually format all output through format function.

dhcole:
:+1:

PR #808, opened 2015-05-22
Make event rsvp selector more specific

### LINKED ISSUE #804: Does not register second RSVP to an event
When I try to RSVP two events at the same time, then the app doesn't register the second RSVP.

Steps to reproduce:
1. Open a group that has more than one event (https://midas-dev.18f.us/projects/11)
2. Toggle the "RSVP" so you're not attending any of the events
3. Refresh the page (optional)
4. Click "RSVP" for the first event
5. Click "RSVP" for the second event
6. Note the reservation was not made for the second event. The app is sending a cancel request instead of RSVP for the second event.

brandtheisey:
The RSVP selector was using a class present on all event attendance buttons (rsvp) to toggle the buttons. So instead of toggling the singular events values when someone RSVP'd or cancelled, it would toggle all of the event buttons, causing unexpected behavior.

--This replaces the vague selector with a more specific selector that only operates on the emitted event's currentTarget (e.currentTarget).

Closes #804
Thanks for the bug report @nhu313!

### DIFF
--- a/assets/js/backbone/apps/events/list/controllers/event_list_controller.js
+++ b/assets/js/backbone/apps/events/list/controllers/event_list_controller.js
@@ -143,9 +143,9 @@ EventList = Backbone.View.extend({
     if (e.preventDefault) e.preventDefault();
     // get the id from the parent event div
     var id = $($(e.currentTarget).parents('div.event')[0]).data('id');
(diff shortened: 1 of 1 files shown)


dhcole:
:+1:

PR #809, opened 2015-05-22
Faster dev builds with npm run watch

dhcole:
Some improvements to the development workflow to speed things up. Now, during development, simply run `npm run watch` to start the application and have it restart on changes.

Removes the grunt-based watch task in favor of `npm run watch`, which uses nodemon. Modifies the default grunt task run on `sails lift` to only build the dev assets, which is faster because it avoids uglifying the files.

:ship: :dash:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,26 @@ We also have a [wiki](https://github.com/18F/midas/wiki) where we keep various d
 
 This project follows the [git flow](http://nvie.com/posts/a-successful-git-branching-model/) branching model of product development.
 
+### Watch command
+
+Use `npm run watch` to automatically regenerate the application after changing files. This command will also source a `.env` file as a bash script if it exists. Use this for setting up environment variables for your development environment, such as:
+
+```sh
+# New Relic settings
+export NEW_RELIC_APP_NAME=midas-local-dev
+export NEW_RELIC_LICENSE_KEY=18956ab3a4d8c772
+
+# LinkedIn auth settings
+export LINKEDIN_CLIENT_ID=77xinrs
+export LINKEDIN_CLIENT_SECRET=11aw6wbHC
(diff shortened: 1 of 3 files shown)


sefk:
When I saw the branch named "dev" I was hoping that someone finally decided to rename devel to dev, which I would be **totally** in favor of. Alas, no.

sefk:
Was there a file not checked in? When I tried to run in my dev environment I got an error:

```
sef@speck (dev) midas> npm run watch

> midas@0.3.4 watch /Users/sef/src/18F/midas
> source .env && nodemon --watch assets --watch api --watch config --watch views -i assets/build

sh: .env: No such file or directory

npm ERR! midas@0.3.4 watch: `source .env && nodemon --watch assets --watch api --watch config --watch views -i assets/build`
npm ERR! Exit status 1
```

dhcole:
@sefk thanks for catching that. The watch script should not fail if the `.env` file is missing. I just fixed that and also added documentation for the `.env` file: https://github.com/18F/midas/pull/809/files#diff-6a3371457528722a734f3c51d9238c13

sefk:
Neato! That works now for me, thanks @dhcole 

Those aren't real keys in your .env file docs I trust. :)

sefk:
Suggestion: instead of the command "npm run watch" how about "npm run dev". watch feels more like the "how" instead of the "why." As as user (in this case, a developer) what I want to do is to run the development stack the recommended way, and I believe this will be our recommended way from now on. 

I could also be convinced that this should just replace "npm run start", if that is only used in a dev context. My guess is when you run in production, you start via "node app.js" directly.

ultrasaurus:
FWIW: "watch" is used by other command line tools I've used to mean the same thing, so it felt familiar

dhcole:
@sefk fake I'm afraid. Use them as you wish :)

dhcole:
Also, I've found "watch" to be a fairly standard term for this kind of thing too, so I'd prefer to stick with convention.

sefk:
ok that makes sense!

PR #810, opened 2015-05-22
Redirect to referring url after oauth login

### LINKED ISSUE #724: When an unauthenticated users tries to volunteer for a task and then logs in with an oauth provider, they should get the volunteer prompt
From @ultrasaurus via #637:

> Also, this flow only pertains to email login, with MyUSA you are directed to /projects (which should only happen if you came in thru the home page) -- feel free to split that out as a separate issue, but for now, this one issue captures the correct behavior above.

There isn't a way to do this currently, since using an oauth provider to sign in redirects the user away from the page, clearing any data in memory.

To do this, we'd need to either set a parameter in the URL to be read after authentication or use a session cookie.

dhcole:
This PR redirects the user to the referring url after logging in with an oauth account. This addresses the issue where logging into the `/people` page redirected away.

It also changes the way the volunteer modal is triggered after log-in so that it works with oauth accounts.

Closes #724

### DIFF
--- a/api/controllers/AuthController.js
+++ b/api/controllers/AuthController.js
@@ -75,7 +75,8 @@ function authenticate (req, res, strategy, json) {
           res.send(userUtils.cleanUser(user, user.id));
         }
         else {
-          res.redirect(sails.config.ui.home.logged_in_path);
+          res.redirect(req.session.logged_in_path || sails.config.ui.home.logged_in_path);
+          delete req.session.logged_in_path;
         }
         return;
       });
@@ -237,7 +238,12 @@ module.exports = {
     if (!target || target === '' || !_.contains(sails.config.auth.config.oauth, target)) {
       return res.send(403, { message: "Unsupported OAuth method." });
     }
-    var config = sails.config.auth.config.config;
+    var config = sails.config.auth.config.config,
+        path = req.headers.referer.split('://')[1].split(req.headers.host)[1];
+
+    // Set referer path for redirection after authentication
+    req.session.logged_in_path = path;
+
     passport.authenticate(target, config[target].params || null)(req, res, function (err) {
       if (err) {
         sails.log.error('Authentication Error:', err);
--- a/assets/js/backbone/apps/tasks/show/controllers/task_show_controller.js
+++ b/assets/js/backbone/apps/tasks/show/controllers/task_show_controller.js
@@ -317,7 +317,10 @@ var TaskShowController = BaseView.extend({
   volunteer: function (e) {
     if (e.preventDefault) e.preventDefault();
     if (!window.cache.currentUser) {
-      window.cache.taskVolunteer = this.model.id;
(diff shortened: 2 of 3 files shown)


api/controllers/AuthController.js
```diff
@@ -237,7 +238,12 @@ module.exports = {
     if (!target || target === '' || !_.contains(sails.config.auth.config.oauth, target)) {
       return res.send(403, { message: "Unsupported OAuth method." });
     }
-    var config = sails.config.auth.config.config;
+    var config = sails.config.auth.config.config,
+        path = req.headers.referer.split('://')[1].split(req.headers.host)[1];
```
sefk: This seems fragile: can there ever be a case where the referrer has a different host than the host header? cnamed hosts maybe?

assets/js/backbone/apps/tasks/show/controllers/task_show_controller.js
```diff
@@ -317,7 +317,10 @@ var TaskShowController = BaseView.extend({
   volunteer: function (e) {
     if (e.preventDefault) e.preventDefault();
     if (!window.cache.currentUser) {
-      window.cache.taskVolunteer = this.model.id;
+      Backbone.history.navigate(window.location.pathname + '?volunteer', {
```
sefk: fragile, what if pathname already has a query string and you'd need to append via & instead. Consider a library not have to worry about special cases.

api/controllers/AuthController.js
dhcole: I don't think there could be a different host than the host header... We need to find the referring header instead of the request path, because we need the source page, not the API call to authenticate.

Do you know of another way to handle this?

assets/js/backbone/apps/tasks/show/controllers/task_show_controller.js
dhcole: The `pathname` property would only contain the actual url path. Existing query params would be available in the `search` property of the `location` object instead. While it's possible that this redirect would overwrite an existing query string, we don't use query strings in this part of the application, and it doesn't really fit with the URL pattern we're using anywhere, so I don't think there would be a conflict. Also, open to suggestions on another approach if you have ideas for this.

screen shot 2015-05-26 at 11 59 38 am

dhcole:
Thanks for the review @sefk. I posted back inline. Generally, I think these issues are mitigated, but I'm open to suggestions / other approaches if there are better ways to go.

ultrasaurus:
Yeah, I agree the query arg approach has its pitfalls, but I think this is a reasonable approach given the constraints.

sefk:
Sorry I didn't reply to your comments @dhcole after my initial review. I
should have been clear that these were just suggestions, not blockers,
since I didn't have a specific error mode in mind. So I'm glad that you
proceeded!

dhcole:
:+1: glad we have this documented too, in case it comes back up in the future.

PR #827, opened 2015-06-04
Log errors in file localization script.

dhcole:
Just a minor update to the migration script for file uploads that adds better logging.


dhcole:
Merging so I can include it in the release I'm preparing now.

PR #842, opened 2015-06-10
[Ready for testing] Notification refactor

### LINKED ISSUE #814: As a developer, it should be easier to add a notification
This issue represents the work to refactor the notification system to be simpler to reflect current usage.  Unused code and configuration should be removed.  

Review wiki page and ensure that it references all the notifications (no more and no less) -- note that there exists a template which appears to be unused:  https://github.com/18F/midas/wiki/Email-Templates

dhcole:
This makes managing email notifications much simpler, forgoing some unused features of the previous configuration.

The notification process now works like this:

1. Create a notification model with an action and a model. The action is a string name for the notification, and the model is a model associated with the notification, for instance, the user for a user welcome email notification. By convention, I'm naming actions as `model`.`method`.`name`, like `user.create.welcome`.
2. When a new notification is created, it triggers a notification email, which builds the data object for the email and constructs the email header and body from templates compiled with the data object.
3. Triggering a notification automatically renders and sends it; `trigger`, `render` and `send` are  all model methods now instead of a separate service.

Adding / editing an email template is much easier. Email templates now consist of two files in a folder named after the action:

```
api/notifications/
  |- user.create.welcome/
    |- notification.js
    |- template.html
  |- layout.html
```

- `notification.js` exports template strings for header fields `to`, `cc`, `bcc`, and `subject` as well as a `data` function that takes the notification's associated model and a callback and returns a `data` object used to build the templates
- `template.html` is an html underscore template file for the body
- `layout.html` is a global layout template that all notification body content gets rendered into before sending

We rely on our our mailing API to inline CSS and convert HTML templates to plain-text.

The basic refactor is set up. Now we need to do the following to migrate over the old data and notifications:

- [x] copy over existing notifications and make sure they are HTML-escaping properly
- [x] replace references to the old notification service (these are commented out now)
- [x] adjust activity feed on the admin page (may need to add in templates for these)
- [x] migrate data from existing notifications and write schema migration script

Resolves #814


dhcole:
This is ready for testing now. I'd like to hold of releasing it until after we merge some of the other PRs that do not require data migrations. But would love help testing. It's a pretty big one!

To test:

```sh
$ npm run migrate                               # migrate data schema
$ node ./tools/postgres/updateNotifications.js  # migrate data values
```

Then you should see notifications for [each notification event](https://github.com/18F/midas/wiki/Email-Templates) as well as an activity feed of recent notifications on the admin dashboard.


cc @brandtheisey @jeremiak

jeremiak:
I've pulled down the branch and have run the migration. I checked out the admin dashboard but received no notifications there nor in my associated email when I created new tasks and comments. Am I testing this correctly?

dhcole:
@jeremiak When you create a new user, new task, volunteer for a task, delete a volunteer from a task, assign a volunteer a task, or comment on a task, Midas should generate an email notification.

If you haven't set up an SMTP provider with Midas, you'll see an error in the console, which for our purposes here is good enough of a test.

To set up an email provider, you can use gmail or a free Mandrill account, with this set in your `local.js` file:

```js
  emailProtocol: 'SMTP',
  smtp: {
    service             : '',
    host                : 'smtp.mandrillapp.com',
    secureConnection    : false,
    port                : 587,
    auth                : {
      user              : '[mandrill user]',
      pass              : '[mandrill pass]'
    },
    ignoreTLS           : false,
    debug               : false,
    maxConnections      : 5
  },
  systemEmail: '[from address (don't use a GSA address — GSA will block the incoming messages]',
```

dhcole:
Also, I thought that was in the docs somewhere, but I couldn't find it, so I added it here: https://github.com/18F/midas/commit/f3f7d0e0df569814508a08e87b94d4e256eed1e3

dhcole:
@brandtheisey last call on this. I'd like to merge it in today. Let me know if you have any issues.

jeremiak:
I see the error/debug log in my console for some of the above specified events so I surmise this works great. Merging now :)

PR #850, opened 2015-06-15
Small installation changes

jeremiak:
Just instead Midas today so here are some notes from my process


dhcole:
Welcome to Midas @jeremiak! :tada: 

:ship:

PR #853, opened 2015-06-17
escape contents of tags

brendanwhitfield:
I believe select2 was supposed to do this transparently, but seems to be disabled (probably because of the HTML emitted by `createSearchChoice()`).


brandtheisey:
Hi @brendanwhitfield is this the issue you are solving with this PR? https://github.com/18F/midas/issues/799

dhcole:
:ship:

Issue #149, opened 2014-03-26
Consistent error messaging for 404 and other error pages.  


polastre:
Apply the Open Opportunities design system pages for 404 and other error pages.  
Design System Link: 

For project or opportunity ids that don't exist anymore, eg `/projects/foobarbaz`, we should show a 404 error page rather than a blank project/opportunity form (which will confuse users).

afturner:
I'll work on this.

Piglady64:
We want to show our own 404 page when a page doesn't exist. Consistent with USAJOBS.

kbwagner:
1/8/20 - This has been completed

PR #855, opened 2015-06-17
added simple 404 responses

brendanwhitfield:
This is a first crack at #149 

When an item isn't found, instead of showing blank project/task/profile pages, this PR uses the `alert_template` to display the error the from the server.

**It turns this:**

before-not-found

**into this:**

not-found-alert


dhcole:
Thanks @brendanwhitfield! We'll review this for the next release!

PR #861, opened 2015-06-19
added filtering policy for user UPDATE requests

brendanwhitfield:
prevents people from arbitrarily turning themselves into admins via the API.

### DIFF
--- a/api/policies/protectAdmin.js
+++ b/api/policies/protectAdmin.js
@@ -1,7 +1,10 @@
 /**
-* Only Admins should be able to set the isAdmin property of the model.
+* Only Admins should be able to UPDATE the isAdmin property of the model.
+* All other references to isAdmin should be scrubbed.
 */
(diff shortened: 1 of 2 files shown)


brandtheisey:
I like this a lot :+1: , people can still be promoted to admin through the correct channel?

dhcole:
:+1: yea, @brandtheisey this would allow admins to promote other admins.

dhcole:
Good catch @brendanwhitfield

brendanwhitfield:
woah, that was fast!

Yes, although the `AdminController` [has it's own handlers](https://github.com/18F/midas/blob/devel/api/controllers/AdminController.js#L242-L260) for setting this flag, and that whole controller is already behind an "admin-only" policy.

dhcole:
We have an internal-only audience so we haven't done as thorough of a review as we should, but this is definitely an important one to lock down.

PR #862, opened 2015-06-19
Prevent users from editting other users via the API

brendanwhitfield:
prevents non-admin users from updating other user's information via the API.

I edited an existing policy to allow admins to perform UPDATEs to arbitrary users. The only other place where this policy was used is [`UserEmailController.findAllByUserId`](https://github.com/18F/midas/blob/devel/config/policies.js#L64) (handled [here](https://github.com/18F/midas/blob/devel/api/controllers/UserEmailController.js#L10-L16)), which I *believe?* is also an acceptable admin action.

### DIFF
--- a/api/policies/user.js
+++ b/api/policies/user.js
@@ -3,8 +3,16 @@
 */
 
 module.exports = function user (req, res, next) {
-  if (req.route.params.id == req.user[0].id) {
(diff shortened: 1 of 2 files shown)


dhcole:
Looks great!

PR #867, opened 2015-06-22
Fixed HTML comments in admin panel with marked

### LINKED ISSUE #865: Markdown / HTML for comments displays in admin activity feed
We're escaping but new rendering the HTML / formatted markdown on the admin activity feed. This means some HTML entities get displayed.

screen shot 2015-06-22 at 9 26 04 am

brendanwhitfield:
Fixes #865 

~~This PR no longer stores comments in the database in escaped form. [This `html()` call](https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/comments/new/views/comment_form_view.js#L133) was returning entities, whereas the `text()` call below was returning raw content.~~ This PR also adds `marked` comment rendering to the admin panel, and uses it to escape the contents, rather than the escaping in `_.template()`. ~~This appears to be backwards compatible with a database full of escaped comments.~~

### DIFF
--- a/assets/js/backbone/apps/admin/views/admin_dashboard_view.js
+++ b/assets/js/backbone/apps/admin/views/admin_dashboard_view.js
@@ -6,6 +6,7 @@ var AdminDashboardTemplate = require('../templates/admin_dashboard_template.html
 var AdminDashboardTable = require('../templates/admin_dashboard_table.html');
 var AdminDashboardActivities = require('../templates/admin_dashboard_activities.html');
 var LoginConfig = require('../../../config/login.json');
+var marked = require('marked');
 
 
 var AdminDashboardView = Backbone.View.extend({
@@ -54,15 +55,22 @@ var AdminDashboardView = Backbone.View.extend({
     _(data).forEach(function(activity) {
 
       if (!activity || ( activity.comment && typeof activity.comment.value == "undefined") ) return;
-      // Strip HTML from comments
+      // Render markdown
       if (activity.comment) {
-        var value = activity.comment.value.replace(/<(?:.|\n)*?>/gm, '');
+        var value = activity.comment.value;
+
+        value = marked(value, { sanitize: false });
+        //render comment in single line by stripping the markdown-generated paragraphs
+        value = value.replace(/<\/?p>/gm, '');
+        value = value.replace(/<br>/gm, '');
+        value = value.trim();
+
         activity.comment.value = value;
       }
       // Format timestamp
       activity.createdAtFormatted = $.timeago(activity.createdAt);
       var template = self.$('#' + activity.type).text(),
-          content = _.template(template, { escape: /\{\{(.+?)\}\}/g })(activity);
+          content = _.template(template, { interpolate: /\{\{(.+?)\}\}/g })(activity);
       self.$('.activity-block .activity-feed').append(content);
     });
 


dhcole:
@brendanwhitfield Looking good. This addresses the issue on the admin feed but creates a new issue on the comment feed on a given post. It breaks the formatting for replies. 

1. Create a comment
2. Click "reply", get HTML formatted block quote
3. Add a replay and post the new comment
4. Formatting is stripped off of the block quote reply

Mind taking a look to see if the escaping can be adjusted to support this?

screen shot 2015-06-22 at 2 53 53 pm
screen shot 2015-06-22 at 2 54 00 pm
screen shot 2015-06-22 at 2 54 05 pm

assets/js/backbone/apps/admin/views/admin_dashboard_view.js
```diff
@@ -54,15 +55,16 @@ var AdminDashboardView = Backbone.View.extend({
...
 
       if (!activity || ( activity.comment && typeof activity.comment.value == "undefined") ) return;
-      // Strip HTML from comments
+      // Render markdown
       if (activity.comment) {
-        var value = activity.comment.value.replace(/<(?:.|\n)*?>/gm, '');
-        activity.comment.value = value;
+        activity.comment.value = marked(activity.comment.value, { sanitize: true });
+        //render comment in single line by stripping the markdown-generated paragraphs
+        activity.comment.value = activity.comment.value.replace(/<\/?p>/gm, '');
```
dhcole: Nice catch for the `<p>` tags

assets/js/backbone/apps/admin/views/admin_dashboard_view.js
```diff
@@ -54,15 +55,16 @@ var AdminDashboardView = Backbone.View.extend({
     _(data).forEach(function(activity) {
 
       if (!activity || ( activity.comment && typeof activity.comment.value == "undefined") ) return;
-      // Strip HTML from comments
+      // Render markdown
       if (activity.comment) {
-        var value = activity.comment.value.replace(/<(?:.|\n)*?>/gm, '');
-        activity.comment.value = value;
+        activity.comment.value = marked(activity.comment.value, { sanitize: true });
```
dhcole: This shouldn't be needed since `santize` is set globally https://github.com/18F/midas/blob/devel/assets/js/backbone/app.js#L24-L27

assets/js/backbone/apps/comments/list/views/comment_item_view.js
```diff
@@ -11,7 +11,7 @@ var CommentItemView = Backbone.View.extend({
 
   render: function () {
     this.model.currentUser = window.cache.currentUser;
-    this.model.valueHtml = marked(Autolinker.link(this.model.value), { sanitize: false });
+    this.model.valueHtml = marked(Autolinker.link(this.model.value), { sanitize: true });
```
dhcole: I'm thinking this was false to allow for the HTML formatting for block quotes. That should be fine as long as the quote text is previously escaped.

brendanwhitfield:
welp, that's embarrassing.

The conundrum came from the fact that we're using HTML *in a place where we don't want HTML*. The comment must exist half-escaped in the database, where the reply formatting can persist. Turns out, the `html()` call that noted originally was *intended* to provide that escape. My mistake, thanks for catching it.

Another possible issue: now that the admin page has markdown rendering, replies can show up has multiline jumbles. Any thoughts on how to handle that?

admin_markdown

dhcole:
I think this is good to go like this. We'll probably do another pass at redesigning the activity feed soon, so we can revisit the line breaks then. Thanks @brendanwhitfield

Issue #290, opened 2014-09-30
As a task creator, I should be able to better specify when it should be completed


gwynnek:
On the Create Opportunity screen, in addition to the "day, days, week, month, few month," add an option with a specific due date.

ultrasaurus:
We need to support:  https://github.com/18F/midas-open-opportunities/issues/35
For specific dates (like April 4, 2015) the workaround of putting text in the description is fine

dhcole:
@ultrasaurus do you have a final list / mock-up of the selection options that should be present? Seems like the lists here and in 18F/midas-open-opportunities#35 are a bit old so I want to make sure I have the latest.

ultrasaurus:
Here's a design which can accomodate the Open Opps time commitments (and I think improves the information architecture in general):
![image](https://cloud.githubusercontent.com/assets/41906/6337653/ff341290-bb5f-11e4-8f51-b0b457e1eb6b.png)

The idea is if someone switches the first popup to "Full-time"  (or 20% in Open Opps) then we would hide "will required [amount of time] of work, and"  since the amount of time is already specified in the first popup.

ultrasaurus:
For open opps, Part-Time / Full-Time would be customized as:
* Quick
* On-going

And:
* a few minutes to 1 hour
* up to 2 hours
* 2-4 hours
* 4-8 hours
* 8-16 hours
* 16-24 hours
* 24-40 hours

must be complete within: [1 week, 1 month, 3 months, 6 months]

from the discussion here:  https://github.com/18F/midas-open-opportunities/issues/35

dhcole:
@jeremiak summing up recent conversations, opportunities are defined by three types of tags, as listed above in @ultrasaurus's comment. To close out this task, we need to 

- [ ] change the existing tags from "part-time", "full-time", and "20%" to  `["one-time", "on-going"]`
- [ ] adjust the duration times to match the above comment
- [ ] adjust the completion time frames to match the above comment
- [ ] restructure the form to read like "This [one-time, on-going] task will require [duration] of work ["each week" if on-going], and must be completed within [completion time frame]

jeremiak:
Quick question, will this require a migration? I just want to make sure I understand how tags work and it seems like the first three items on your comment @dhcole require a db migration. Am I correctly understanding how the tags work per the updated the wiki page?

dhcole:
Not sure exactly. @ultrasaurus should we migrate these tags from the old ones to match the new ones in your comment above?

[...]

PR #872, opened 2015-06-23
Better tags for task time lengths

jeremiak:
This PR changes the copy on task creating/editing views to reflect a few new tags related to task length. Specifically this should address #290 as it changes the tags loaded by default (upon running `npm run init`) and also the copy for the task views.

An open question I still have is if we want to move the old tags over to the new names, and what we should do in the cases where there isn't a clear mapping. For instance, should we map "20%" time to "on going?"


dhcole:
@jeremiak This is looking good. Per conversations in stand up, are you taking a look at a data update script to move old tags to the new ones in this PR?

jeremiak:
@dhcole as soon as we finalize the copy changes per #290 this should be good for :eyes:. The end result is a data migration script that is idempotent and does a three distinct tasks:

0. Renames tags that have a direct mapping
0. Adds new tags that were not present before
0. Migrates associated users, projects, and tasks off of old tags and onto new ones. It then deletes the old tags from the database.

jeremiak:
I believe this is ready for testing as I followed @GailSwanson's advice per #290

dhcole:
Looks good. 

*Developer note*: run `node tools/postgres/migrateOldTags.js` to migrate to the new tags.

PR #875, opened 2015-06-23
added admin policy to TagEntity controller

brendanwhitfield:
Prevents non-admin users from modifying existing tags in the database. [It was noted](https://github.com/18F/midas/blob/devel/api/controllers/TagEntityController.js#L11-L15) that this was intended to be secured, and only accessible by admins.

### DIFF
--- a/config/policies.js
+++ b/config/policies.js
@@ -147,7 +147,9 @@ module.exports.policies = {
 
   TagEntityController : {
     // Purely for administrative functions
-    '*': 'authenticated'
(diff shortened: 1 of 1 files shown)


dhcole:
This works since it still allows any authenticated user to create new `TagEntity` models. The comment you linked to from the controller is actually out of date. Since the [tag refactor]() we do allow authenticated users create new `TagEntities` in addition to admins; however there's no reason for them to update or delete the models. :+1:

PR #878, opened 2015-06-24
fixed ownerOrAdmin policy for volunteer destroy

brendanwhitfield:
Plugs another API hole where any user could delete volunteers. The `ownerOrAdmin` policy was already present, but it passes requests that [aren't tasks or projects](https://github.com/18F/midas/blob/devel/api/policies/ownerOrAdmin.js#L11-L14), and the `isOwner` flag was never set. This PR adds a `volunteer` policy that sets the `isOwner` flag, and removes the project/task clause from `ownerOrAdmin`.

**Question**
Would it be better to remove `ownerOrAdmin` in favor of individual `ownsTask`, `ownsProject`, policies? The actual ID checks (to set the isOwner flag) are endpoint-specific, so it might be safer to have this done in one policy, rather than require a dependent policy to set flags.

### DIFF
--- a/api/policies/ownerOrAdmin.js
+++ b/api/policies/ownerOrAdmin.js
@@ -8,10 +8,6 @@ module.exports = function addUserId (req, res, next) {
   if (req.isOwner || req.user[0].isAdmin) {
     return next();
   }
-  // if this isn't a project or task request, pass through
-  if (_.isUndefined(req.proj) && _.isUndefined(req.task)) {
-    return next();
-  }
   // Otherwise not allowed.
   return res.send(403, { message: "Not authorized." });
 };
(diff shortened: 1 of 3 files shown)


dhcole:
@brendanwhitfield We want task creators to be able to delete volunteers from their own tasks as well. This is done before moving a task to the "assigned" state — the task creator will pare down the volunteer list to the needed quantity.

Do you mind adjusting this to check for whether the requesting user owns the task to which the volunteer is assigned as well?

As for whether they're separate policies, what do you think makes the most sense? Since the policies cascade, I don't have an strong opinions on which way we go.

brendanwhitfield:
Your right, other policies cascade in a similar manner anyways. It's probably better/safer to change less code.

I'll update the PR when I get some time. Thanks for catching that!

brendanwhitfield:
Updated

dhcole:
Works well. Thanks @brendanwhitfield

PR #880, opened 2015-06-25
Download files with the correct filename, rather than file ID

brendanwhitfield:
simply adds the `download` attribute to the link, so that files aren't downloaded with names like `2`, and `38`.


dhcole:
Nice. :ship:

PR #884, opened 2015-06-25
Fixed ajax error for world data in Firefox

brendanwhitfield:
In Firefox and Chrome, the D3 ajax request on the `/people/` page for `world-110m.json` can fail. This is because it requests

`http://localhost:1337/people/data/world-110m.json`

instead of,

`http://localhost:1337/data/world-110m.json`

This only happens when there is a trailing slash in the page URL (`http://localhost:1337/people/`). The solution was simply to make the path relative to the root, which now operates correctly. Tested in Firefox 38 and Chrome 43.

**To reproduce**
- point Firefox or Chrome at `http://localhost:1337/people/`. *(Note the trailing slash)*
- The map will fail to load, and the network inspector will show a 404'd request for `world-110m.json`

### DIFF
--- a/assets/js/backbone/apps/people/controllers/people_map_controller.js
+++ b/assets/js/backbone/apps/people/controllers/people_map_controller.js
@@ -28,7 +28,7 @@ PeopleMap.Controller = BaseController.extend({
       })
     });
     gatherData.push(function (cb) {
-      d3.json("data/world-110m.json", function (err, world) {
+      d3.json("/data/world-110m.json", function (err, world) {
         if (err) {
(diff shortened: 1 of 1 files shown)


jeremiak:
Cool. @brendanwhitfield how do I reproduce the bug on a `devel` based branch and consequently test it on this branch? Which view makes the D3 request?

brendanwhitfield:
@jeremiak Sorry, looks like I was working faster than I was documenting! Issue updated.

dhcole:
Makes sense. Good fix.

PR #889, opened 2015-06-30
Remove obsolete comment about TagEntity controller

dhcole:
Since the tag refactor, we do allow authenticated users to create new TagEntities

Follows up on #875


sefk:
:+1:

brendanwhitfield:
:+1: Thanks

Issue #700, opened 2015-04-09
As a new member, I should be prompted to fill out my profile after I sign up


ultrasaurus:
Next step after signing up.  Show another dialog, with
* agency (required)
* location (required)
* A brief public biography (optional) with large text area (like in profile)

[Go to My Profile] [Browse Tasks]

related:
https://github.com/18F/midas/issues/464 Prompt people to fill out their profile, if they haven't already, when they add a task or project
https://github.com/18F/midas/issues/463 Prompt people to fill out their profile, if they haven't already, when they volunteer

dhcole:
Per sprint-planning conversation, we discussed refactoring the sign-up process into this:

![img_5137](https://cloud.githubusercontent.com/assets/170641/8167687/eb8b973e-136d-11e5-8af3-be462b47680c.JPG)

This would kick in at any point a user tries to create a new account. It's a series of modal screens where the account is not actually created until the user makes it through both steps. Users can start with email and password or oauth, but they both then need to be validated as having a federal email address.

Name, email, password, and agency are required fields.

@brandtheisey how will this affect your environment? I'm thinking since all of your users are automatically signed in, they would bypass this whole process.

@jeremiak I'll work on the API-level components of this, like verifying an email is on the federal domain list and returning a matching agency name for a given email domain. The rest of the work should all take place at the backbone level.

dhcole:
@ultrasaurus please verify my memory here on the flow when you have a moment.

ultrasaurus:
Looks right to me, except we talked about an optional text area for bio. Th

dhcole:
Thanks, I forgot that.

jeremiak:
Just an update on this issue before the weekend, working off of the `agency-location-signup` branch.

I have added an agency and location config options to `login.json` and when either are present the registration form becomes two forms. The first sub-view is identical to the current view, except instead of 'Submit' the positive button says 'Next'. This is because there is a secondary sub-view that contains the registration options that are configured on.

The registration works fine and there is client side validation that all the present forms are properly filled out. However, the API doesn't accept the extra parameters of location and agency, which are `TagEntity` IDs. Will work on that next and then making sure to update/expand the tests.

PR #890, opened 2015-06-30
Slight refactor of the TagFactory

### LINKED ISSUE #881: Tag selection prompts to recreate existing tags
During tag selection, if an existing tag is found, the "Click to create new tag" choice is still present and functional. This is confusing (given the matches shown below), and can even result in duplicate tags in the database.

existing-tags

brendanwhitfield:
Frontend changes only

I was about ready to start adding new tag fields to the profile page for my client, when I noticed a lot of redundant code regarding Tag dropdowns. This PR helps keep things DRY by using the TagFactory where complicated `select2` calls were before. To do this, a handful of new options and existing code was added to the TagFactory.

80% of the TagFactory code you see being changed is actually just code moving from place to place. There were various versions of the same chunks of `select2` code floating around inside Midas. I have gathered them into the TagFactory, and provided extra options where neccessary. The remaining calls to `select2` are either trivial, or transact with endpoints other than `/api/ac/tag` (for instance, the `task_edit_form_view` uses `/api/ac/user` for setting task ownership). My hope was to make the TagFactory a bit better at its job.

## Change summary
- New options for `tagFactory.createTagDropDown` (also added a doc comment to explain it all):
  - `multiple`: whether or not to allow multiple tags
  - `allowCreate`: while most inputs allow for new TagEntity creation, a few do not, such as the `Agency` field.
  - `data`: sets the initial data of the tag selector
- `tagFactory.createTagDropDown` now returns the constructed `select2` jQuery element, selected by `options.selector`. 
- Correctly handles location tags in "single mode" (`multiple=false`), allowing it to be used on the Profile page.
- When a tag already exists, do *not* prompt to "create new" (fixes #881)
- Replaced redundant `select2` calls with `tagFactory.createTagDropDown({ ... })`. (Also plugs a few lingering sanitization holes by sharing common escapement code). Pages where this happened:
  - `/profile/:id` agency, location
  - `/admin/tag` agency, skill, topic
- Updated existing `tagFactory.createTagDropDown({ ... })` calls to make use of the new options.
  - `/tasks/:id` skill, topic, location
  - `tag_show_view.js` agency, skill, topic, location (so, pretty much everywhere...)
- Removed an old [`createDiff`](https://github.com/18F/midas/blob/devel/assets/js/backbone/components/tag_factory.js#L154-L187) function that was no longer used. Please correct me if I'm wrong. My search for calls to this function came up null, and it seems to assemble tag structures that aren't used anymore. Instead of being diffed, tags are simply sent in bulk with model updates, with existing tags referenced by ID (see [here](https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/profiles/show/views/profile_show_view.js#L469-L479) for the current way this is done).
- Removed old ["administrative" tag code](https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/tag/show/views/tag_show_view.js#L90-L133). The `#input-tags` selector actually returns empty, and silently fails anyway. I could not find reference to `input-tags` or any of those events being broadcast (`<target>:tag:new` and `<target>:tag:save`). I also saw no functional difference, but again, please correct me if I'm wrong.
- The Profile page now correctly registers `change` events on the Tag dropdowns.

I'd really appreciate a thorough review on this one, since I've been rearranging some grey matter. Midas *should* be functionally identical to the way it was before, with the exception of the fix for #881, and the patching of a few escapement problems.

Tested in Firefox 38 and Chrome 43

### DIFF
--- a/assets/js/backbone/apps/admin/views/admin_tag_view.js
+++ b/assets/js/backbone/apps/admin/views/admin_tag_view.js
@@ -37,44 +37,16 @@ var AdminTagView = Backbone.View.extend({
 
   tagSelector: function(type) {
     var self = this;
-    $('#' + type).select2({
-      placeholder: 'Search for a tag',
-      minimumInputLength: 2,
-      formatResult: function (obj, container, query) {
-        return obj.name;
-      },
-      formatSelection: function (obj, container, query) {
-        return obj.name;
-      },
-      createSearchChoice: function (term, values) {
-        var vals = values.map(function(value) {
-          return value.value.toLowerCase();
-        });
-        return (vals.indexOf(term.toLowerCase()) >=0) ? false : {
-          unmatched: true,
-          tagType: type,
-          id: term,
-          value: term,
-          name: "<b>"+term+"</b> <i>click to create a new tag with this value</i>"
-        };
-      },
-      ajax: {
-        url: '/api/ac/tag',
-        dataType: 'json',
-        data: function (term) {
-          return {
-            type: type,
-            q: term
-          };
-        },
-        results: function (data) {
-          return { results: data };
-        }
-      }
-    }).on('change', function(e) {
+
+    var $sel = this.tagFactory.createTagDropDown({
+      type: type,
+      selector: "#" + type,
+    })
+
+    $sel.on('change', function(e) {
       var $el = self.$(e.currentTarget);
       self.tagFactory.addTagEntities(e.added, self, function() {
-        $('#' + type).select2('data', null);
+        $sel.select2('data', null);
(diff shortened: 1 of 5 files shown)


jeremiak:
@brendanwhitfield thanks for this PR, in addition to DRYing out the code a bit it helped me to understand how to use the `tagFactory` component in #700.

I ran the tests locally and it looks like they all pass. I also manually tested the application and I didn't spot any regressions.

LGTM :+1:

jeremiak:
Per today's standup, I'm going to merge this down so I can base the changes I've got going for #700

PR #891, opened 2015-06-30
Added more console help output to the tagtool

brendanwhitfield:
Made the tagtool a little more user friendly.

Replaces failures like this (caused by no arguments):

```
/home/brendan/midas/tools/tagtool/tagtool.js:15
if (tagType.length === 0) {
           ^
TypeError: Cannot read property 'length' of undefined
```

with output like this:

```
This is a tool for loading tags into the database in bulk.

Usage:
    ./tagtools.js [type] [file]

    type - The string type of tags to be loaded
    file - file containing a newline delimited list of tag values to be loaded
```

Also changed the permissions to make it executable by the user.

### DIFF
--- a/tools/tagtool/tagtool.js
+++ b/tools/tagtool/tagtool.js
@@ -1,21 +1,40 @@
 #! /usr/bin/env node
 
+function help_and_quit(msg) {
+  if(msg) {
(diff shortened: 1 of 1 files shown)


dhcole:
:+1:

PR #892, opened 2015-07-01
Prevent users from updating comments they don't own

brendanwhitfield:
Previously, comment UPDATE requests were processed without question. With this policy, only comment owners, and admins are allowed to alter existing comments.


dhcole:
:+1:

PR #896, opened 2015-07-04
Add script to build monthly report

dhcole:
Script with template and instructions for building a monthly report of closed pull requests.


jeremiak:
From my perspective, this looks good to me. I was able to use the included documentation and `report.js` and was able to easily and successfully generate a `generated.md`.

Is there an issue or something more specific to help determine when this PR is done? If not, I can only surmise that its supposed to generate a simple report and it seems to do that -> :+1:

dhcole:
Thanks for the review. This is just a one-off script for generating reports for OCSIT.

PR #902, opened 2015-07-09
Authentication refactor

dhcole:
Work in progress. Just about down. Needs a few more things:

- [x] remove unused code (user service, user email, user auth user password controllers and models)
- [x] verify email is from federal domain for authentication
- [x] update tests


dhcole:
Tracking an upstream issue: https://github.com/18F/myusa/issues/712

jeremiak:
Testing this PR out now, just want to make sure I understand what the functionality should be. 

@dhcole after I migrate and set the proper `local.js` var the registration is restricted to federal domain emails. But what should happen to users who had accounts without this domain restriction? Is this check just for new accounts?

dhcole:
Right, the check is only for new accounts. I plan to make a list of existing accounts that violate the rule and send to Lisa for manual review.

jeremiak:
This PR looks great! It works as intended and when the proper config variable is set restricts sign up to a whitelist of Federal domains based on the supplied email address.

In addition, it really simplifies the code for authentication and looks like it hands most of that responsibility off to Passport? I really like that we're getting rid of models like user password :)

dhcole:
:+1: Thanks for the review! I'll finish up the tests and resolve the merge conflicts, then we can :ship:

dhcole:
This is ready expect for the issue @sefk noticed in #904.

It looks like something changed with Travis's environment were falsy values are printing differently...

```
"user_id","name","username","title","agency","location","bio","admin","disabled"\n1,"Admin User","admin@midascrowd.com",,,,,true,\n2,"Default User","tester1@midascrowd.com",,,,,,\n3,"Password Reset User","testreset@midascrowd.com",,,,,,\n
"user_id","name","username","title","agency","location","bio","admin","disabled"\n1,"Admin User","admin@midascrowd.com","","","","",true,false\n2,"Default User","tester1@midascrowd.com","","","","",false,false\n3,"Password Reset User","testreset@midascrowd.com","","","","",false,false

```

dhcole:
@sefk @jeremiak the test failure is due to a minor update in `json2csv`. I pinned to a specific version in this branch so that should fix the issue.

sefk:
Nice debugging @dhcole. Glad to see those tests passing again. Once this lands I'll merge.

dhcole:
Going to merge this in and test on staging with more real data. 

## Development notes

This includes a migration schema that adds a new model table as well as migrates local and oauth users to the new passport models. `npm run migrate` should set this up.

Also this pins `json2csv` to a version, so you may need to run `npm install` if you have a more recent version that is causing export tests to fail.

@brandtheisey I have no way of testing the SSPI authentication. I *think* it works, but we should do some thorough testing in your environment before you deploy this release.

dhcole:
## Additional developer note

**Before applying this update** you need to verify that your database does not have duplicate accounts. Do this by running the SQL query in `./tools/postgres/removeDuplicateUsers.sql`. If you do have duplicate accounts, you can remove them by changing this query -- see the comments in the file.

PR #904, opened 2015-07-11
Browse people

sefk:
This is a work-in-progress pull request. While this isn't ready to pull yet, it's at a stage where I'd like to get some feedback.  @dhcole or @brandtheisey or @jeremiak -- what do you think about the code; @GailSwanson and @ultrasaurus what about the UX?

Current status (from commit b2d1fa0)

    Move people search into browse

    Instead of having people search in its own hidden endpoint wire into
    normal "browse" feature.

    - Search or autocomplete not wired up yet.

    - Routing to /profiles currently doesn't work. If you go home and then
      select the "people" tab that works, but then if you hit refresh it
      doesn't (404). Something is broken in /profiles routing.

    - Currently loading map data geo data directly from the map view. Not
      only bad for separation of concerns, but loading every page view is
      dumb, wasteful, and slow.

    - Paginate the peoplist list when you get lots of results.

    - Push search terms into navigation to make robust for refresh,
      sharable/bookmarkable, etc.

Note there are also some vestiges of my first attempt at adding navbar support. 

browse-people


sefk:
Fixed the big outstanding problems:

- /profiles external routing works now
- map now cleanly interacts with other search terms (thanks @jeremiak)
- geo data loading now cleaner and more straightforward (thanks @jeremiak)

I believe this is now ready to demo and get some initial feedback and testing.

sefk:
I made the one must-fix change from yesterday's demo, removing "browse" from all the navbar elements. I believe this is ready for review now, hence removing "WIP" from the title.

I did try playing with updating the map based on the search result but it causes a bunch of interaction oddities that will require more time and thought to get right, so I'm deferring that for now. Check out the branch ```browse-people-map-update``` with 5fe54a5 if you want to play with that. It's just the naive three-line change.

Looking forward to some review feedback @dhcole and/or @jeremiak. Thanks!

sefk:
Failing test in Travis appears to be unrelated (task / user export) and breaking in PR #908 also.

api/controllers/SearchController.js
```diff
@@ -49,6 +50,7 @@ function search(target, req, res) {
       }
     });
 
+    // TODO: needs escaping to prevent SQL injection
```
dhcole: The `pg` library does this: https://github.com/brianc/node-postgres/wiki/FAQ#8-does-node-postgres-handle-sql-injection

sefk: Right, I should have said "parameterized queries" than doing the escaping ourselves. Indeed that's how I handled it in my own code, I just put this TODO in here as a reminder to do that elsewhere. I guess I could update the comment.

assets/js/backbone/apps/browse/browse_app.js
```diff
@@ -13,25 +13,24 @@ var TaskShowController = require('../tasks/show/controllers/task_show_controller
...
 var HomeController = require('../home/controllers/home_controller');
-var PeopleController = require('../people/controllers/people_main_controller');
 
 
 var BrowseRouter = Backbone.Router.extend({
 
   routes: {
     ''                          : 'showHome',
-    'projects(/)'               : 'listProjects',
+    'projects(/)(?:queryStr)'   : 'listProjects',
```
sefk: these are vestiges of my earlier attempts at pushing search terms into the navbar, and then capturing / using them later (bookmarkable searches, refresh, etc.) This doesn't do any harm, but not useful either without anything to use queryStr. I think its worth leaving but I'm happy to remove too.

api/controllers/SearchController.js
dhcole: Gotcha -- these queries aren't from your work. No problem. I think we just fix them separately.

dhcole:
Finished a code review and this like great! Thanks for the major lift @sefk 

Also confirmed tests pass when merged with `devel`.

Just one usability issue I ran into that I'd like @ultrasaurus and @GailSwanson to weigh in on:

When clicking the People tab while not logged in, you get this message:

screen shot 2015-07-17 at 1 55 34 pm

This is because profiles are protected behind authentication.

Working groups and Opportunities are not behind authentication. 

Is this OK as is or should we handle this interaction differently before merging?

dhcole:
Per @GailSwanson in standup, we could hide this tab until users are logged in so that users are not confused.

sefk:
Thanks for the review @dhcole and for highlighting that one auth issue. I discussed with @GailSwanson and implemented in 57991fa. Ready to go? :smile:

dhcole:
:rocket:

sefk:
woot woot

PR #909, opened 2015-07-17
Add script for reviewing and deleting duplicate users

dhcole:
In theory, users shouldn't have been able to create multiple accounts with the same email address. An unreproducible bug allowed this to happen (#548)

With #902, we now enforce a database constraint on the uniqueness of `midas_user.username`. Before that update can be applied, we need to remove an duplicates that would violate the constraint.

This script produces a list of duplicate accounts to be deleted. Once that list has been reviewed, changing `SELECT *` to `DELETE` in the beginning of the query will remove those records.

Modified from https://wiki.postgresql.org/wiki/Deleting_duplicates

It prioritizes keeping enabled accounts and those with recent logins.


jeremiak:
:+1: :white_check_mark:

PR #910, opened 2015-07-17
Catch for an error when finding tasks with users without tags

dhcole:
Just a small patch for something I saw coming up on the staging server.


dhcole:
Need this to fix an issue on `midas-dev`, so merging now.

PR #911, opened 2015-07-17
Add from name to outgoing emails. Closes #866

### LINKED ISSUE #866: Emails should have a from name
So instead of seeing noreply@... in the email client, we see the name.

dhcole:


jeremiak:
:+1:

PR #913, opened 2015-07-17
Fix demo script. Simplify init script

dhcole:
Uses Sails directly instead of over http, avoiding the need for a temporary admin account.

Fixes the demo script. Note, this script still throws an error at the end but it works for getting the data set. Not worth troubleshooting the error as this should be rewritten too like the init script.


jeremiak:
LGTM!

dhcole:
Thanks for the quick look!

PR #914, opened 2015-07-18
Auth update

dhcole:
While checking out domain validation against the list of domains in production, I realized two things:

1. Our list does not allow for subdomains, such as `example.gsa.gov` when only `gsa.gov` is on the list.
2. Our list does not include `.mil` domains. They are not in any of the lists of domains to which we have access. Assuming all domains ending in  `.mil` are federal, this allows them access


jeremiak:
Code makes sense and manually tested signup for a user with a proper .gov domain (and one with a subdomain) and was successful. my attempt to signup with a personal email was rejected.

On another note, would it be helpful to have another column on waffle for something like "Ready for review"? That way we can distinguish between PRs that are ready and those that are still in progress. Or we can just operate under the assumption that when a PR is opened on Github its ready for review?

dhcole:
@jeremiak thanks for the review. 

For this PR, I was planning to leave it open to catch any feedback that comes up, like the last commit to change the error message for blocked accounts.

In general, I like to submit PRs early and before they're finished for transparency and discussion around what I'm doing. I think a "Ready for review" tag is a great idea. I'll create it and we can set up a column in waffle for it.

PR #917, opened 2015-07-20
New users require location and agency

### LINKED ISSUE #700: As a new member, I should be prompted to fill out my profile after I sign up

jeremiak:
This PR is not yet ready for review (and will be properly tagged when it is). This should require that new users who sign up with the `local` strategy (aka email and password) also supply their agency and location during sign up.

Resolves #700


jeremiak:
I think the last bit I need to do before its ready for review is to document where one sets these configuration variables.

I’m thinking we should start a wiki page since we’ve added 3 in the last week. Any thoughts about using a wiki versus the `README.md`?

jeremiak:
Have a related open PR on [midas-cookbook](https://github.com/18F/midas-cookbook/pull/36) to include these new variables in the production environment.

jeremiak:
Documenting configuration options in a new wiki page: https://github.com/18F/midas/wiki/Deploying-Midas-on-your-own

api/services/protocols/local.js
```diff
@@ -40,6 +43,22 @@ exports.register = function (req, res, next) {
     return next(new Error('No password was entered.'));
   }
 
+  if (sails.config.requireAgency) {
+    if (_.isUndefined(req.param('agency'))) {
+      req.flash('error', 'Error.Passport.Password.Missing');
+      return next(new Error('No agency was entered.'));
+    }
+    agencyID = parseInt(req.param('agency').id);
```
dhcole: Generally a good idea to specify the [radix](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Parameters) explicitly, so how about `parseInt(req.param('agency').id, 10)`? Or you could go with just `+req.param('agency').id`

api/services/protocols/local.js
```diff
@@ -40,6 +43,22 @@ exports.register = function (req, res, next) {
...
+  }
+
+  if (sails.config.requireLocation) {
+    if (_.isUndefined(req.param('location'))) {
+      req.flash('error', 'Error.Passport.Password.Missing');
+      return next(new Error('No location was entered.'));
+    }
+    locationID = parseInt(req.param('location').id);
```
dhcole: Same as https://github.com/18F/midas/pull/917/files#r35770180

assets/js/backbone/apps/login/templates/login_template.html
```diff
@@ -104,16 +132,32 @@
...
+        <button type="button" id="register-next" class="btn btn-c2">Next</button>
+      </div>
+    </div>
+  </div>
+
+  <div class="row" id="registration-footer-prev-submit">
     <div class="col-sm-12">
       <div class="pull-right">
+        <% if (login.agency.enabled === true || login.location.enabled === true) { %>
+        <button type="button" id="register-previous" class="btn btn-c0"><span class="badge error" style="display:none;">!</span> Prev</button>
```
dhcole: @jeremiak why show the error badge on the previous button? Should this be on the submit button?

jeremiak: @dhcole my thought was that since the registration is spread over two different views you would want some indicator that the error is on the previous view and therefore know where to rectify the situation.

It was not the most thought out so I'm open to suggestions, it could just be on the submit button and then the user would check both screens.

api/services/protocols/local.js
jeremiak: I included `parseInt` because when I add the tags on [L74-79](https://github.com/18F/midas/blob/agency-signup-new-auth/api/services/protocols/local.js#L74-L79) I wanted an integer and not a string as it expects an ID. I can try it again without `parseInt` and if I need it I will add a radix of `10`

assets/js/backbone/apps/login/templates/login_template.html
dhcole: Ok. Fine for now. We can revisit as part of the larger redesign.

jeremiak:
Looks great, passes all automated tests and worked via manual testing :ship:

PR #919, opened 2015-07-22
Sort time tags in proper order

### LINKED ISSUE #916: Sort tags in a way that makes sense for users
The new task duration tags look great, but they're sorted by database ID:

screen shot 2015-07-20 at 12 38 42 pm
screen shot 2015-07-20 at 12 38 27 pm

Could we sort by their logical sequence? One possible solution is to manually sort them in the migration script and set their `createdAt` timestamp to (pseudo) ` now + i`, then alter the front end code to sort them by `createdAt`.

jeremiak:
The autocorrect controller generally returns tags sorted by `id` which is great for most of our use cases. However, it was problematic with the `task-time-estimate` tags as the `id` isn't the best indicator of order for those tags.

This PR sorts `tagEntities` returned through auto-correct by the `updatedAt` field instead of `id`. It also updates a previous migration script (`/tools/postgres/migrateOldTags.js`) to make sure to delete old tags and touch the remaining `task-time-estimate` tags so that the `updatedAt` order is the proper order for UI display.

To close #916

### DIFF
--- a/config/settings/sources.js
+++ b/config/settings/sources.js
@@ -56,7 +56,8 @@ console.log('Loading... ', __filename);
       ],
       include: [
         'name',
-        'type'
+        'type',
+        'updatedAt'
       ]
     },
     'wikipedia': {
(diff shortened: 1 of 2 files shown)


dhcole:
:+1:

PR #924, opened 2015-07-29
Add scripts for importing tasks from Wordpress site

### LINKED ISSUE #847: As a user or admin, I should be able to see 2014 data
As an admin, I should be able to see FY14 vs FY15 data that is accurate
As a user, I should see all of the work I have done in my profile

import from tasks from WordPress

dhcole:
This imports legacy task data from the Wordpress-based Open Opportunities site.

All imported tasks will be associated with user 1. We should change this to Lisa's profile ID as we run it on each environment.

All published, assigned, and completed timestamps are set to the date the task was created.

The associated user and timestamps can be changed by editing the task.

This gives us our best take on getting old data into the system for the admin metrics. Resolves #847 

## Running the script

First, download the export xml from the Wordpress site. Then run

```
$ ruby tools/wordpress-import/convert_xml.rb tools/wordpress-import/digitalservicesinnovationcenter.wordpress.2015-07-27.xml > tools/wordpress-import/import.json
```

to process and convert the xml to a JSON format that can be imported. (h/t to @ultrasaurus for the original version of this script).

Then run 

```
node tools/wordpress-import/import-tasks.js
```

to import the JSON tasks into the database.

After running this, we probably should delete `import.json` to slim down the repository size. It's included here now so we don't need to share the full xml file or run `convert_xml.rb` in each environment.

### DIFF
--- a/tools/wordpress-import/convert_xml.rb
+++ b/tools/wordpress-import/convert_xml.rb
@@ -1,0 +1,69 @@
+#!/usr/bin/ruby
+
+# Install gem with `$ gem install reverse_markdown`
+# Run with `$ ruby context_xml.rb [path/to/input_file.xml] > [output_file.json]`
+
+require 'nokogiri'
+require 'reverse_markdown'
+require 'time'
+require 'json'
+
+filename = ARGV[0]
+
+items = []
+File.open(filename) do |file|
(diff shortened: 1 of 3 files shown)


jeremiak:
:rocket:

Issue #922, opened 2015-07-27
add 18F Google Analytics Account 


gbinal:
Per the [18F Analytics Standards](https://github.com/18F/analytics-standards), can we please integrate https://openopps.digitalgov.gov/ into the 18F Google Analytics (GA) account?  

To do so, you'd need to insert [this template](https://github.com/18F/analytics-standards#javascript-code-snippet) and use this UA code - `UA-48605964-17`.

gbinal:
And to https://pages.18f.gov/midas/ as well.

jeremiak:
It appears that the Open Opportunity app already is configured for both DAP and 18F analytics - [see config here](https://github.com/18F/midas-open-opportunities/blob/master/config/settings/analytics.js#L4-L6).

However, looking at the 18F analytics dashboard makes me think that it isn't configured correctly

jeremiak:
Further update. @gbinal we have the following currently being served on Open Opportunities but the 18F GA account doesn't seem to reflect any new data since June 3 (related issue: #918). Any ideas or obviously wrong things with the snippet below?

```
    <script>
      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
      })(window,document,'script','//www.google-analytics.com/analytics.js','ga');

     ga('create', 'UA-48605964-17', 'auto');
     ga('send', 'pageview');
   </script>
```

dhcole:
@jeremiak, I think you've made some progress on this over slack, right? Anything we should update here?

jeremiak:
@dhcole the update from Slack is that I'm to implement the new 18F provided snippet. I'll get that PR updated by COB today, should be really straight forward

jeremiak:
Made some changes recommended by @gbinal in Slack in #946 that amount to:

0. Force to use `https://` for the collect call back to GA
0. `anonymizeIp` and `forceSSL` are enabled

PR #927, opened 2015-07-29
Include 18F analytics guild snippets

jeremiak:
In #922 @gbinal requested the addition of 18F Google Analytics snippets to the Open Opportunities application as well as the 18f-pages midas site.

This PR is for the static site hosted with 18F-pages only. It adds the supplied analytics snippet to the `index.html` page as well as the `/developer/index.html` page.

Per the issue, both use `UA-48605964-17` as their UA code.

### DIFF
--- a/developer/index.html
+++ b/developer/index.html
@@ -537,5 +537,17 @@ function load() {
 
 load();
 </script>
+
+<script>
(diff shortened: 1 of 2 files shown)


dhcole:
:+1:

PR #929, opened 2015-08-03
Sort migration scripts numerically

brendanwhitfield:
Ran into a lexical sorting problem when I added my tenth migration script, so I simply added the numeric sorting flag.


dhcole:
Yea, good call

PR #931, opened 2015-08-04
Admin Export Task Data Report

### LINKED ISSUE #718: As an admin, when I export task data, I would like to see additional data
Would like to have task id, task state, agency of task creator, completion date, archived date -- ideally the state would use the configured text.

Current attributes are:
* project_id
* name
* description
* created_date
* published_date
* assigned_date
* creator_name

awhig:
The CSV download file of the admin Tasks Report now has
additional columns as requested in the issue report.

Resolves #718

### DIFF
--- a/api/models/Task.js
+++ b/api/models/Task.js
@@ -54,14 +54,13 @@ module.exports = {
     'description': {field: 'description', filter: exportUtils.nullToEmptyString},
     'created_date': {field: 'createdAt', filter: exportUtils.excelDateFormat},
     'published_date': {field: 'publishedAt', filter: exportUtils.excelDateFormat},
-    'assigned_date': {field: 'createdAt', filter: exportUtils.excelDateFormat},
+    'assigned_date': {field: 'assignedAt', filter: exportUtils.excelDateFormat},
     'creator_name': {field: 'creator_name', filter: exportUtils.nullToEmptyString},
     'signups': 'signups',
     'task_id': 'id',
     'task_state': 'state',
     'agency_name': {field: 'agency_name', filter: exportUtils.nullToEmptyString},
-    'completion_date': {field: 'completedAt', filter: exportUtils.excelDateFormat},
-    'archived_date': {field: 'deletedAt', filter: exportUtils.excelDateFormat}
+    'completion_date': {field: 'completedAt', filter: exportUtils.excelDateFormat}
   },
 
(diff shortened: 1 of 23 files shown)


dhcole:
Thanks @awhig! Great to have your help on this. Will review tomorrow and get tagged for the next release.

api/models/Task.js
```diff
@@ -55,7 +55,12 @@ module.exports = {
     'published_date': {field: 'publishedAt', filter: exportUtils.excelDateFormat},
     'assigned_date': {field: 'createdAt', filter: exportUtils.excelDateFormat},
     'creator_name': {field: 'creator_name', filter: exportUtils.nullToEmptyString},
-    'signups': 'signups'
+    'signups': 'signups',
+    'task_id': 'id',
+    'task_state': 'state',
+    'agency_name': {field: 'agency_name', filter: exportUtils.nullToEmptyString},
+    'completion_date': {field: 'completedAt', filter: exportUtils.excelDateFormat},
+    'archived_date': {field: 'deletedAt', filter: exportUtils.excelDateFormat}
```
dhcole: @awhig nice work. Jus two things here on dates:

Looks like this should be `assigned_date` from `task.assignedAt` according to #718 

Also, tasks with empty dates show up as `Invalid date` in the CSV. We should add a test here for whether the date is valid: https://github.com/18F/midas/blob/devel/api/services/utils/export.js#L53

@awhig do you mind updating the PR with these? Thanks!

awhig: Hi @dhcole,

According to #718 she says:

> Would like to have task id, task state, agency of task creator, completion date, **archived date** > -- ideally the state would use the configured text.

Not sure what you want to change here.


I'll make the change for the testing the date is not null.

awhig: Is that a previous bug, using task.createdAt instead of task.assignedAt for assigned_date? If so, I can make that fix as well.

dhcole: Ah, I see. I was confused by that and the bulleted list in the ticket. Yes, `assigned_date` should use `assignedAt`, we do not actually store an archived date — archived is different than deleted, since archived is a state. So how about we fix `assigned_date` and remove `archived_date `?

awhig: I'll make these changes.

dhcole:
:+1: thanks @awhig. Works well.

PR #932, opened 2015-08-04
Single page create opportunity view

### LINKED ISSUE #318: As a task creator, I need better prompts (Rework the new opportunity workflow)
img_20141006_204051

img_20141006_204101

cc: @ultrasaurus

jeremiak:
This should close #318 when merged. The idea is to redesign the task creation flow so that it is a bit easier to comprehend and complete. We want to achieve this by removing the modal concept from this flow and having a single, short form on a URL-linkable page that enables users to create opportunities.

In addition, we wanted to remove the madlib style text and introduce more conventional form controls and inputs.

This PR will be ready to merge when:
- [x] Migration script to add a `completeBy` timestamp column to the tasks table
- [ ] Current `time-frequency-estimate` tags are renamed so that they can be used as frequency estimates. For example, "1 week" will be changed to be "Once a week"


Gruntfile.js
```diff
@@ -82,7 +82,8 @@ module.exports = function (grunt) {
         ext: '.min.css',
         extDot: 'last',
         options: {
-          report: 'min'
+          report: 'min',
+          sourceMap: true
```
dhcole: How much does this add to the file size? Should we commit this, or just make it as a local change during development. Or, perhaps it could be toggled based on `process.env.NODE_ENV !== 'production'`

dhcole:
Looks great! We can follow up with the issue view page before we tag the release. :ship:

jeremiak:
Hey @dhcole I made a commit and push to this branch before I realized that you had merged it down. Sorry about that. Would you prefer I open a new branch/PR or can we just merge this one down?

My last commit should take care of the "Deadline" tag in that it should only be shown on tasks with an explicit completion date.

PR #935, opened 2015-08-05
Define node and npm versions during builds

dhcole:
Travis and our EC2 servers started failing on builds today. It looks like something is no longer compatible with npm v1.4.28. It doesn't seem to be related to new code, because the `master` branch, which passed a test a few days ago now fails without any changes.

A work-around is to just make sure builds use a more recent version of npm. This PR makes sure Travis updates npm before installing the application. And [this](https://github.com/18F/midas-cookbook/pull/37) makes sure the build server does the same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ sudo: false
 language: node_js
 node_js:
   - "0.10"
+before_install:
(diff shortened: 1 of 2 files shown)


jeremiak:
LGTM

Issue #469, opened 2015-01-05
when I go to my profile, I should see the whole profile URL


ultrasaurus:
When I go to my profile, the URL shows:  https://midas.18f.us/profile
but it is the same as https://midas.18f.us/profile/3  (which is what is included in the email when I share my profile) 
It should show the full URL, so I know what the URL is if I want to tell other people without using the share button

polastre:
You could, instead, make the profile URL something like `/profile/polastre` so that it is more human readable.  Just an idea.

ultrasaurus:
@polastre if we want to change the URL, that would be a separate issue.  This issue represents the fact that I already have a full URL, I just can't see it unless I share my profile with someone via the share button.

PR #936, opened 2015-08-06
Added user id to end of profile link

awhig:
User wanted the logged in user's profile
link to contain the user id, so copy/
pasting of this link from browser
location bar would work

https://github.com/18F/midas/issues/469


jeremiak:
Looks great @awhig, thanks for this and for the corresponding test update :)

PR #937, opened 2015-08-06
Attachment policy

### LINKED ISSUE #883: Restrict file uploads to task creator & participants
Currently, any user can upload a file to any task. Seems like we should lock this down to task creators and admins. @ultrasaurus what do you think?

cc @NoahKunin

dhcole:
This restricts file uploads to only users who have been verified in some way:

- admins
- task creators of open or assigned tasks
- task participants of assigned tasks
- project owners (not used for Open Opportunities)
- all signed in users can upload files that have no attachment (like profile photos)

The API will reject any other file upload or attachment requests. The front end selectively hides the "add attachment" button.

Tests for the API should pass.

Resolves #883

### DIFF
--- a/api/policies/attachment.js
+++ b/api/policies/attachment.js
@@ -1,0 +1,47 @@
+module.exports = function(req, res, next) {
+
+  // Admins can do anything
+  if (req.user.isAdmin) return next();
+
+  // Attachments require a task or project
+  if (req.param('taskId')) {
+
+    // Find task to check permissions
+    Task.findOne({ id: req.param('taskId') }).exec(function(err, task) {
+      if (err || !task) return res.badRequest('Error uploading file.');
+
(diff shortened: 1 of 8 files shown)


jeremiak:
@dhcole A task only gets into the "Assigned" state by an admin, but once it is in such a state, any user can easily click "Participate" and then upload files to that task without any intervention from the admins or task creator. Is this the expected behavior?

dhcole:
That's a good call. I don't think we need to worry about that in this case, but it's a good follow up issue to remove the participate option from tasks that are assigned.

jeremiak:
Great, this code works as advertised. I think a weird bug (that isn't resulting from this code change, but currently exists) is that any task participant can delete attachments and logged that as #939.

Going to merge this down now.

Issue #939, opened 2015-08-07
Any logged in user can delete any attachments on any task


jeremiak:
Reviewing #937 made me aware of a bug where any logged in user can delete any attachment on any task. To replicate:

0. Find a task (or make one) that has an attachment
0. Log in as another user who is not an admin nor a participant with the task
0. You will be able to delete attachments associated with that task

dhcole:
Closed by #940

PR #940, opened 2015-08-07
Only allow owners or admins to delete attachments

jeremiak:
Per #939 any logged in users could delete any attachment on a task/opportunity.

This PR will restrict the ability to delete attachments to admin users and the users who are owners of that attachment.


dhcole:
Added a fix for unauthenticated users. I just ran into the same issue with #937. This should be good to go now once tests pass.

Issue #832, opened 2015-06-04
As a member, when browsing (profiles, tasks, projects) I want the URL to reflect my current search


sefk:
If a select a city I get a list of people in that city. But I can't share that link, or hit refresh, or bookmark. If the URL reflected my selection then I get all that stuff naturally.

Thought of this based on @ultrasaurus's comments on #830, which should have the same URL behavior.

sefk:
Now that the profile search (people) has been folded into the rest of "browse" alongside of projects (working groups) and tasks (opportunities) this should be generalized to capture search terms for all browse searches. 

It might make sense to hold this off until the search rewrite / cleanup is done, or at least designed.

dhcole:
Let's resolve this as part of #863

PR #945, opened 2015-08-11
Make tags links to search results

### LINKED ISSUE #863: Tags listed on task, project, and profile pages should link to searches
Via Janelle during testing:

> Is there a way to make those topic areas/skills to the left of the task clickable? I would be interested in clicking on the topic to see what other tasks are available under that topic?

To me, this seems like the missing link that will make tags actually useful to the end user.

jeremiak:
In order to close out #863 I made a few additions to how the list views work (also satisfies #832). With this PR the following functionality should be changed:

0. All "skill" tags should be links. This will be present on the individual opportunity/task and the individual project/working group view as well as the collection views. This is to directly close #863.
0. The link works by passing along the desired tag name in a query parameter to the respective collection. For instance, if you wanted to find all tasks that have been tagged with "javascript" you would navigate to `/tasks?search=javascript`. If you wanted "javascript" and "html" it would be `/tasks?search=javascript+html`.
0. Clearing the search filter criteria should set the URL back to the "clean" state (i.e. `/tasks`)

@dhcole this should be ready for your review when you've got some time

### DIFF
--- a/assets/js/backbone/apps/browse/controllers/browse_list_controller.js
+++ b/assets/js/backbone/apps/browse/controllers/browse_list_controller.js
@@ -29,7 +29,7 @@ Browse.ListController = BaseController.extend({
   initialize: function ( options ) {
     // this.options = options;
     this.target = options.target;
-    this.queryParams = options.queryParams || {};
+    this.queryParams = options.queryParams || false;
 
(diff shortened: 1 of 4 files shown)


dhcole:
Looks great @jeremiak :ship:

PR #946, opened 2015-08-11
New Google Analytics snippet

jeremiak:
Per #922 we want to update the Google Analytics snippet with a [new one from the 18F analytics guild](https://github.com/18F/analytics-standards#javascript-code-snippet). The most observable differences are:

0. Force to use `https://` for the collect call
0. `anonymizeIp` and `forceSSL` are enabled.


dhcole:
:+1:

Issue #737, opened 2015-04-24
As a member, when I add a skill or topic on my profile, it is confusing to see the "none" tag


ultrasaurus:
I think we should just hide "None" as soon as someone adds one tag.

![image](https://cloud.githubusercontent.com/assets/41906/7325455/254cbf1c-ea73-11e4-9548-07b82dfaed61.png)

save then see:
![image](https://cloud.githubusercontent.com/assets/41906/7325467/359766ec-ea73-11e4-87d2-8d3f8ab58dc2.png)

When I add a new one, the UI makes more sense, since I can also delete existing tags:
![image](https://cloud.githubusercontent.com/assets/41906/7325483/537b8814-ea73-11e4-9c8d-63977b9952d9.png)

awhig:
I think it might be better, when in edit profile mode, the yellow tags are not displayed at all. The tags are only shown and manipulated in the edit box. Yellow tags are then only shown in read only mode.

This would be less confusing (deleting current tags happens in the same area that you add new tags), but also takes care of the original "none" tag issue.

dhcole:
@awhig, so have all the tags, current and new, in the input field? That makes sense to me. Good idea.

awhig:
I made the change in the PR. All tags, new and old, are now shown in the edit box. The yellow tags are not shown in edit mode.
<img width="409" alt="screen shot 2015-08-11 at 10 17 24 am" src="https://cloud.githubusercontent.com/assets/6233406/9204468/468060fe-4012-11e5-9c9f-f11cb9511699.png">

PR #948, opened 2015-08-11
New and old tags are edited in input box

awhig:
When in edit mode, all tags are shown in the input
box.

https://github.com/18F/midas/issues/737


jeremiak:
Looks great to me, thanks much @awhig!

Issue #800, opened 2015-05-19
Need to refresh to see uploaded profile image


nhu313:
When I upload an image, I can't see it until I refresh the page.

Steps to reproduce:
2. Click on your name, then "Profile"
3. Click "Edit Profile" link on the top right corner
4. Click the camera to upload a photo
5. Watch the network requests
6. Choose a photo
7. Note the progress bar display, but the photo is not displayed
8. Note the network request shows the photo is uploaded and retrieved, yet it is not displayed (perhaps something is wrong with the api?)
9. Manually refresh the page (Cmd + R)
10. Note only the small profile image on top of the right corner is changed
11. Manually refresh the page again (Cmd + R)
12. Note the big profile picture is changed

Note: I signed up using my email.

brandtheisey:
Are you seeing this behavior on https://midas-dev.18f.us/ ? 

What browser and OS are you using?

I just tested this with Chrome from a Mac and cannot duplicate the behavior.

nhu313:
Yes, I'm testing it on (https://midas-dev.18f.us/)[https://midas-dev.18f.us/]. I have Firefox 38.0. I tried it on Chrome and it works fine. Maybe this is firefox problem. Wonder if they're caching something.

dhcole:
#helpwanted for more FireFox testing / debugging.

awhig:
I can repro this on Mac OS X, Firefox 40.

PR #949, opened 2015-08-12
Fix issue with profile image update in FF

awhig:
Firefox seems to be caching the first image and not
retrieving the new image when the  CSS is bumped.
Probably because the URL to the image never changes
(for example /api/user/photo/13).
My solution was to put a timestamp as a URL parameter when
retrieving the new image, (ie. /api/user/photo/13?1439337665153).
I believe this busts FF image cache, and forces FF to
get the newly uploaded image.

https://github.com/18F/midas/issues/800

### DIFF
--- a/assets/js/backbone/apps/profiles/show/views/profile_show_view.js
+++ b/assets/js/backbone/apps/profiles/show/views/profile_show_view.js
@@ -202,7 +202,8 @@ var ProfileShowView = Backbone.View.extend({
   updatePhoto: function () {
     var self = this;
     this.model.on("profile:updatedPhoto", function (data) {
-      var url = '/api/user/photo/' + data.attributes.id;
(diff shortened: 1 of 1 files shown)


jeremiak:
@awhig Thanks for just killing it with all of these pull requests. A quick favor: could you tag things with the "Ready for Review" label once the PRs are in the state you want (they might be already). It just helps me see what I should be looking at in a glance :)

awhig:
Hi @jeremiak 

I don't think I can add any labels. I don't have write access to the repository.

jeremiak:
@awhig Oops, my bad. No worries then, I'll just review the PRs as they come in :D

jeremiak:
Clever, @awhig. Thanks! :ship:

Issue #944, opened 2015-08-10
Long links break the page layout for tasks


dhcole:
Issue here: https://github.com/18F/midas-open-opportunities/issues/70

awhig:
I added overflow:hidden to the div.

The link is now clipped. User probably should then add link text, so the display is better.

<img width="629" alt="screen shot 2015-08-11 at 6 26 41 pm" src="https://cloud.githubusercontent.com/assets/6233406/9214840/a08595c0-4057-11e5-9c76-f53fc6799164.png">

jeremiak:
Instead of `overflow: hidden;` we will use `word-break: break-all;`. Thanks again @awhig!

PR #950, opened 2015-08-12
Fix for long link breaks task template

awhig:
Added overflow: hidden to div.task-show-description

https://github.com/18F/midas/issues/944

### DIFF
--- a/assets/styles/application.css
+++ b/assets/styles/application.css
@@ -1468,7 +1468,7 @@ ul.task-tags > li {
   font-size: 18px;
   margin-bottom: 10px;
   line-height: 1.5em;
-  overflow: hidden;
+  word-break: break-all;
 }
 
 .task-edit-description {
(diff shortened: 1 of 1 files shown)


assets/styles/application.css
```diff
@@ -1468,6 +1468,7 @@ ul.task-tags > li {
   font-size: 18px;
   margin-bottom: 10px;
   line-height: 1.5em;
+  overflow: hidden;
```
jeremiak: Hey @awhig instead of using overflow, let's use `word-break: break-all;` which is [widely supported](http://caniuse.com/#search=word-break) so that the link is entirely visible to the end user and not clipped.

awhig: Done.

jeremiak: Cool, thanks for the quick turnaround! I'll pull it down and just give it quick glance before merging.

jeremiak:
@awhig thanks so much for your contributions!

PR #953, opened 2015-08-13
More time intervals for task metrics

### LINKED ISSUE #905: as an admin, I would like to see metrics by quarter
Current it is by fiscal year... 

image

dhcole:
This adds support on the admin dashboard for several additional time intervals, in addition to Fiscal Year, as previously supported:

screen shot 2015-08-13 at 12 56 07 pm

screen shot 2015-08-13 at 12 57 13 pm

The `undefined` column are tasks with missing dates. Since we started tracking `completedAt`, `publishedAt` and `assignedAt` after the initial launch of the platform, some tasks do not have these dates defined.

Resolves #905

### DIFF
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -246,46 +246,94 @@ module.exports = {
    * with the above endpoint.
    */
   taskMetrics: function(req, res) {
-    var getFY = function(input) {
+    var pad = function(num, size) {
+          size = size || 2;
+          var s = num + '';
+          while (s.length < size) s = '0' + s;
+          return s;
+        },
+        getFY = function(input) {
           if (!input) return undefined;
           var date = new Date(input);
           return new Date(date.setMonth(date.getMonth() + 3)).getFullYear();
         },
-        currentFY = getFY(new Date()),
-        fiscalYears = _.range(2014, currentFY + 1),
+        getYear = function(input) {
+          if (!input) return undefined;
+          return new Date(input).getFullYear();
+        },
+        getMonth = function(input) {
+          if (!input) return undefined;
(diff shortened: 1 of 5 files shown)


dhcole:
Also includes an upgrade to the module we're using for storing session data that prunes the session table automatically. It's quite big on production, so this should help.

jeremiak:
This looks pretty good, a few quick issues I've noticed:

0. The last column/time section under any of the views will have a weird heading. For instance, on the Quarterly view it reads "QNaN", on Weekly it will be "WNaN", and Yearly will result in "undefined".
0. The weekly view seems to skip some weeks (http://up.jeremiak.com/cFhQ), is this because there isn't data for those weeks on my local instance?

dhcole:
> The last column/time section under any of the views will have a weird heading. For instance, on the Quarterly view it reads "QNaN", on Weekly it will be "WNaN", and Yearly will result in "undefined".

These are tasks that don't have the associated date. I should change the label to me more clear.

> The weekly view seems to skip some weeks (http://up.jeremiak.com/cFhQ), is this because there isn't data for those weeks on my local instance?

Yes. Do you think it would be more clear to include all weeks and fill empty ones with zeros?

dhcole:
@jeremiak I relabeled the undefined values. I'm inclined not to address the missing weeks that are empty at this time  — I don't see a quick fix for it.

jeremiak:
"No date" as a label is much clearer to the user.

No, I think its pretty clear that those weeks have no data (just wanted to confirm). In addition, if there were more columns it would make the crowded horizontal scroll a bit harder to read. Going to merge this now.

PR #954, opened 2015-08-13
Copy cleanup on task create page

dhcole:
Add links, consistent punctuation, adjust alignment:

screencapture-localhost-1337-tasks-new-1439487679632


dhcole:
@GailSwanson let me know if anything else should change before shipping. This is the only thing blocking @LisaLNelson's testing.

dhcole:
@GailSwanson per standup, this is ready for your review. Let me know what you think / what else should get done before sending to Lisa.

dhcole:
From slack:

> “X business days" should be “Two business days”

Issue #797, opened 2015-05-18
All user query (new /api/user/all endpoint) slow on dev


sefk:
Look at the "network" tab on chrome developer tools when looking at the [people page](https://midas-dev.18f.us/people) on staging. You see time to first byte (TTFB) for "all" in the 600ms - 1000ms range. Only 15KB of data is being returned!

This isn't what I see in my dev environment, of course. 60ms to return 19KB. But that's localhost, quiet postgres server, artificial (demo) data.

To debug I would either need better test data (see #712) or access to staging.

sefk:
While I can't say for sure, it's very likely that the problem could be the Sails ORM doing a dumb query. [This gist is the Postgres log][pglog] when hitting ```/api/user/all``` on my dev environment (13 users each with a tag or two). 

  [pglog]: https://gist.github.com/sefk/800236501cfd6990f78d

While I haven't profiled this, line 6 of the gist certainly looks to be a potential culprit. Instead of an inner join, the sql generated does a subquery for every foreign key and then union's them all together. You could imagine that a query planner *might* be smart enough to notice and optimize this case but not only is that unlikely, but also unnecessary. 

Three options for proceeding 

1. **Waterline Fix.** I've searched the Waterline and Sails lists and there are several issues filed that may relate to this, but not sure. I'm wary of opening my own issue until I discuss with @dhcole and likely characterize this using their own schema / tests. My reading however makes me believe that this isn't some fix already made somewhere, not a quick & easy fix.

2. **Hand Join.** User.find(), then do TagEntity.find() and join ourselves. Ugly to have to do this but a) two queries would be much more efficient, b) it's more data btw app server and DB, but we can winnow that down before sending to the client, and c) it would still work with the file-based backend used for testing

3. **Hand query().** We can always write the sql ourselves and call query() on it, but that won't work on a non-sql backend, like one that we use while running unit tests. 

Personally I think the "hand join" option is what I'd lean toward, but I wouldn't do it until I had the same problem repro'ed in my own dev environment (see #712 again).

Note there's issue [#944](https://github.com/balderdashy/waterline/issues/944) that may be related but may not. That looks to be a memory issue at larger data sizes than we're seeing.

dhcole:
I can confirm I *do* see this on the production DB as well. It takes about 1000ms to respond to this endpoint.

It seems like running the query is relatively fast, but the node.js code to build the query is the big bottleneck. 

[...]

PR #967, opened 2015-08-19
Add populateFast hook for fast many-to-many associations

dhcole:
This conditionally replaces the standard `Model.populate()` method with one that's up to ***11 times*** faster. It generates a much simpler SQL query for retrieving associated data from many-to-many relationships.

This solution is implemented as a hook, so it could be spun out into it's own npm module for use on any Sails application. It's generic and gets all of its settings from the standard `populate` method.

To use, set `populateFast: true` on your connection's configuration:

 `config/connections.js:`
 ```js
module.exports = {
   postgresql: {
     adapter: 'sails-postgresql',
     populateFast: true
   }
};
```

Since it uses a SQL query, it will only work with adapters that support SQL.

Here are two (previously) slow pages:

## Before

screen shot 2015-08-19 at 6 30 06 pm

screen shot 2015-08-19 at 6 28 58 pm

## After

screen shot 2015-08-19 at 6 27 22 pm

screen shot 2015-08-19 at 6 27 47 pm

Refs #797 
Refs balderdashy/waterline#944

h/t to @sefk for doing the investigative work into why populated queries were so slow.

For sake of comparison, here are the original (standard Sails / waterline) and new queries: https://gist.github.com/dhcole/5455c7723ef968820c2a

### DIFF
--- a/api/hooks/populateFast/index.js
+++ b/api/hooks/populateFast/index.js
@@ -1,0 +1,113 @@
+/*
+ * A faster way to populate many-to-many associations
+ *
+ * Optionally overrides Model.populate() with a faster query
+ * for manty-to-many associations. This only works on SQL-based adapters.
+ *
+ * To use, set `populateFast: true` on your connection's configuration:
+ *
+ * `config/connections.js:`
+ * module.exports = {
+ *   postgresql: {
+ *     adapter: 'sails-postgresql',
+ *     populateFast: true
+ *   }
+ * };
+ *
+ */
+
+var Deferred = require('sails/node_modules/waterline/lib/waterline/query/deferred'),
+    normalize = require('sails/node_modules/waterline/lib/waterline/utils/normalize');
+
+// Format from https://github.com/rhpwn/sails-hook-populate-fields
+module.exports = function(sails) {
+
+  function patch() {
+    Deferred.prototype.populateOriginal = Deferred.prototype.populate;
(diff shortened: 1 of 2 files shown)


jeremiak:
LGTM, that's a really impressive performance improvement! Nice work @dhcole and @sefk

mikermcneil:
Love it- @particlebanana and I just talked about it, and this would be a perfect addition in waterline-sequel.  The reason there's such a huge UNION query is because it's the catchall for every possible combination of where/limit/skip/sort when populating.  That said, there's no reason we can't introspect the `criteria` dictionary, determine that there is a possible optimization, then take advantage of it here (https://github.com/balderdashy/waterline-sequel/blob/master/sequel/lib/criteriaProcessor.js#L79)

mikermcneil:
cc @tjwebb

tjwebb:
@dhcole Wow, awesome work.

I've also made an improvement for Postgres that uses their `array_agg` function to speed up joins: https://github.com/waterlinejs/postgresql-adapter/blob/master/lib/util.js#L304-L324. Similar strategy: make one query instead of several. (the `array_agg` magic is here: https://github.com/waterlinejs/postgresql-adapter/blob/master/lib/util.js#L399-L403).

I like the idea of using `populateFast` if available. One of the issues that's still limiting the performance of my query above is that Waterline still calls `find()` first, even though it doesn't need to.

PR #968, opened 2015-08-22
Adds DAP reporting to 18F pages part of the app

konklone:
This adds DAP reporting to the 18F pages part of the app. I'm not sure how to modify the Sails settings for the DAP source for the `devel` branch.


dhcole:
Thanks @konklone

PR #970, opened 2015-08-24
Bug fix for missing volunteer tags

dhcole:


dhcole:
Simple bug fix. Merging after tests pass so we can get the next release staged.

Issue #798, opened 2015-05-18
People view: add zoom and pan to map


sefk:
Current world-scale map will be difficult for many use cases
- us-centric team -- then the rest of the world is irrelevant
- key parts of the team in a tight cluster, i.e. US Northeast. In that case selecting and/or hover will be difficult
- viewing anything on a small screen

Would be nice to zoom by either +/- buttons when viewing on a normal screen, and a pinch gesture when on a phone/tablet.

austinkeeley:
It looks like the current map is just a static vector graphic. We could swap this out with a real GIS map. I have some experience with Leaflet and OpenLayers. I could build out a prototype and try to have it maintain the existing look and feel of the basemap.

This could also open up some interesting features, like being able to search for opportunities within a bounding polygon.

ultrasaurus:
the current map is built with d3, it's just not dynamic (yet!).  Have you experience with the library?  agree there are lots of great possibilities here

ultrasaurus:
this depends on https://github.com/openopps/openopps-platform/issues/1546

Piglady64:
We will do usability testing to see if users want this feature. Moved to 1546. Closing

PR #978, opened 2015-08-28
Adding pan and zoom controls to profile map

austinkeeley:
This is change related to issue https://github.com/18F/midas/issues/798 to add pan and zoom controls to the profile map.  I swapped out the D3 map with a Leaflet map while trying to maintain the same style of the underlying map and circle markers.

Because Leaflet doesn't natively support TopoJSON, I had to generate GeoJSON using data from the Natural Earth website in order to get a vector basemap of the countries. All of the Natural Earth vector data are released in the public domain.

image


dhcole:
@austinkeeley this is great! Thanks for the contribution. We'll review it this week. 

@sefk you may be interested in this since it builds on your earlier work, albeit with a different library.

dhcole:
@austinkeeley thanks again for the PR. I like using Leaflet for this. I'm seeing the map expand outside of its contain and it the zoom handlers break after zooming in:

screen shot 2015-09-16 at 1 30 56 pm

Do you see this as well? I have Leaflet 0.7.5 installed.

dhcole:
Seems like maybe there is some missing CSS, perhaps from Leaflet?

Issue #979, opened 2015-08-31
Improve naming and visual representation of badges


GailSwanson:
As a participant, I would like to see badges displayed with the new names and graphics across the platform. 

*badges*
"Agency Rep" keep name, new graphic
"Founder"  keep name, new graphic
"Tapathon"  keep name, new graphic
"Tapas" gets relabeled "Connection" with new graphic (puzzle piece image)
"20%" gets relabeled "Journey" with new graphic (bicycle)

GailSwanson:
All formerly awarded badges should be converted. 

In a future iteration, we plan to redo the manner in which we award badges so that they are graduated rather than awarding duplicates to an individual. When that is implemented, people will retain their historical badges on their profiles even if those badges are deprecated.

jeremiak:
What are the definitions of those badges? That is to say, how does a user get awarded these various badges?

Is this documented somewhere from the old platform?

GailSwanson:
I don't know if they were documented anywhere, however, the definitions are not complex.

Completing a one time task gets you a Connection badge
Completing an ongoing task gets you a Journey badge
Participants in the Tapathon got a Tapathon badge
There is only one Agency Rep recipient. We will redefine if we decide to award any future ones.
Founder badges will only be those who have already been awarded those badges.

dhcole:
Migrating old badges that we don't plan to support poses a few issues:

1. How do we know who gets those badges? We'd have to manually assign them. What if those users don't have accounts on the new system yet?
2. This would break our current approach, which is that badges would be dynamically generated based on querying data in the database. If these are not going to apply to users who qualify for these conditions in the future, how to we make the distinction?

Have you thought about how we should address these? Can we put historic badges aside for another iteration and focus on new ones since only new badges would apply to the wall of fame feed?

GailSwanson:
Is it not possible to rename/label the old badges that have historically been awarded and continue to award with the new names according to the established algorithm? I'm not recommending changing the algorithm at this point in time, just the names and graphics

GailSwanson:
The tapathon and founders badges won't have additional/future users who meeting those qualifications, so we should be fine on those. The agency rep badge will most likely be in the same boat. 

If/when we decide to change the badging algorithm, we will have to figure out how to maintain the historical badges. That's the big reason I didn't want to up end that part of things with this iteration.

dhcole:
[...]

PR #981, opened 2015-09-01
Wall of Fame implementation

### LINKED ISSUE #871: Create a Wall of Fame
The current Open Opportunities [wall of fame](http://gsablogs.gsa.gov/dsic/category/wall-of-fame/) should be implemented in Midas.

Also, this should trigger a notification email to the user: https://github.com/18F/midas/issues/819

jeremiak:
This PR will close #871 and #979 when merged. It implements the concept of the "Wall of Fame" by altering the logged in home page to feeds of the most recent activity on the platform.

Behavior changes:

0. When a user is logged **out** they will not be able to see this view. Instead, they will see our traditional homepage.
0. When a user is logged **in** they will not be able to see the traditional homepage, and instead see this dashboard.

Some development tasks we still need to do to fully enable this interface:
- [ ] Return user objects with agency and/or location if it exists (returning all tags would be fine)
- [x] Implement badging queries or notifications (#979)


dhcole:
@jeremiak added the controller for the endpoints needed for this. Check the `ActivityController` for the available routes, such as `/api/activity/badges`.

There are a few todos in the comments to finish up, but the format of the API  responses should be stable.

jeremiak:
@dhcole this is exactly what I needed, thanks!

dhcole:
@jeremiak updated the `network` and `count` endpoints to support the standard `where={...}` param, like `http://localhost:1337/api/activity/network` and `http://localhost:1337/api/activity/network?where={%22completedAt%22:%20{%20%22%3E%22:%20%222015-06-01%22}}`

jeremiak:
@dhcole thanks for that enhancement. I can update the activity collection to pass that through so that the closed opportunity count can be from the last 12 months.

@GailSwanson this should now be in a place where its ready for a design critique

jeremiak:
Just finished up the design feedback implementation with @GailSwanson, so I think this is ready for code review/testing @dhcole

dhcole:
@jeremiak: Great work!

I pushed a few small changes to make sure we're escaping content using a different endpoint for the images so we get the default image for users without photos.

screen shot 2015-09-16 at 12 38 07 pm

If you're good with this, then we're ready to :ship:

jeremiak:
@dhcole thanks for those, especially the default profile picture! I'll go ahead and merge

PR #995, opened 2015-09-17
Fix theme collision with new dashboard

jeremiak:
The theme was clashing with our new dashboard to make it look like this:

The culprit was the `.home` class that was still being applied to the dashboard container. I removed the class from the view's render code to ensure that the class is only used for the unauthenticated page.

I wasn't able to get the themes to properly apply locally but I was able to test the removal of this class with Chrome dev tools to yield this:

@dhcole I believe this is ready for review.


dhcole:
:+1:

Issue #990, opened 2015-09-14
As an Open Opportunities member, I need the ability to view full time details according to agency (GSA)


GailSwanson:
This could be accomplished in an extensible way through agency and task type search filters.

Issue #989, opened 2015-09-14
As a GSA employee, I need the ability to express interest in a full time detail opportunity


GailSwanson:
First iteration of this feature might not restrict non-GSA users from viewing or selecting the opportunity, but at minimum must inform all users that only GSA are eligible.

PR #997, opened 2015-09-18
Full time detail option for GSA

### LINKED ISSUE #988: As a GSA employee, need the ability to post full-time detail opportunities
Include information unique to the detail task type such as:
-Link to application process and process information
-Restrictions on who is eligible to apply (i.e.GSA only)

jeremiak:
This pull request should be the first iteration of the full time detail feature. Changes to the application are as follows:

0. There is a script to add a "Full time detail" tag that is restricted to GSA affiliated users
0. A task creator who is also a GSA affiliated user can create tasks with this new tag
0. GSA affiliated users looking for tasks through the regular search interface will see these opportunities, but users that don't have GSA as their agency will not be able to see these tasks.

@GailSwanson ready for your review

To close #988 #989 #990

### DIFF
--- a/api/services/passport.js
+++ b/api/services/passport.js
@@ -412,7 +412,10 @@ passport.serializeUser(function (user, next) {
 });
 
 passport.deserializeUser(function (id, next) {
-  User.findOne(id, next);
+  User.findOne(id).populate('tags').exec(function(err, user){
+    if (err) next(err);
+    next(null, user);
+  });
 });
 
 module.exports = passport;
--- a/assets/js/backbone/apps/browse/views/browse_main_view.js
+++ b/assets/js/backbone/apps/browse/views/browse_main_view.js
@@ -238,9 +238,13 @@ var BrowseMainView = Backbone.View.extend({
       // projects and tasks get tiles
       filteredCollection = _.filter(filteredCollection, function (item) {
         // filter out tasks that are full time details with other agencies
-        var userAgency = _.where(window.cache.currentUser.tags, { type: 'agency' })[0],
+        var userAgency = { id: false },
             timeRequiredTag = _.where(item.tags, { type: 'task-time-required'})[0];
             fullTimeTag     = false;
+
+        if (window.cache.currentUser) {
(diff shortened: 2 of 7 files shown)


ultrasaurus:
@jeremiak haven't looked at the code, just want to make sure that if someone at GSA sends someone else at GSA a direct link to the task, it doesn't require login... is that how it works?

jeremiak:
@ultrasaurus Yup! Anybody can send anybody else a link for a task (https://openopps.digitalgov.gov/tasks/387 for example) and they can view it without authentication.

GailSwanson:
We should make sure that doesn't open any back door entry points to unauthenticated users viewing PII, such as the task creator's profile.

ultrasaurus:
@GailSwanson I don't expect that to happen since this is old behavior, but would be good to make sure we have automated tasks for that.

FYI: for government use of Midas, there is no PII in the system.  All of our information as federal employees is a matter of public record -- we require the use of work email address and locations are work locations.  However, we don't want to open up the system to spammers and want to support private sector / non-gov use cases, so your point is absolutely valid, just not technically a PII concern.

dhcole:
This pr does not change profile visibility. And issues with profile visibility should be addresses separately.

dhcole:
What I mean by that is that all this PR does is hide tasks in search. All the rest of how tasks or profiles are visible remains the same. There should not be any issues there, but if there are, researching and addressing them are out of scope this this PR

jeremiak:
I should've added this to the original PR but this does require a script be run to add the new tag for Full Time details at GSA. From the project root, you should run:
`node tools/postgres/addGSAFullTimeDetail.js` and then start the server normally

dhcole:
@jeremiak I updated with two small changes. The biggest one is that if you try to create a task after logging in, you got an error, because the profile returned from the authentication endpoint didn't populate tags. I fixed this by adding in an additional request to get the user's full profile.

I'll merge once tests pass.

GailSwanson:
It took a chunk of time for @jeremiak and I to get issues ironed out so that I could view the PR. All good now.

Issue #830, opened 2015-06-04
As a member, when I look at the map, I would like to be able to click on an agency and see the people in that agency


ultrasaurus:
This should change the URL, so people can bookmark a map or send the link to other people via email or whatever

ultrasaurus:
depends on turning the map back on: https://github.com/openopps/openopps-platform/issues/1546

kbwagner:
No longer applicable

PR #1001, opened 2015-09-22
Rewrite search in browse views

### LINKED ISSUE #947: Larger search issues might require a refactor
We wanted to break out #863 into two parts, one to represent the smaller task of ensuring that tags are linkable and one to represent the larger issues around the search mechanism.

PR #945 addresses the smaller, more concrete task, but here are some links to issues that were used to document the larger search problem.

0. Search for groups filters uses "and" while opportunities uses "or" - #801 
0. Search autocomplete should not show duplicate tags - #585 
0. Search should AND and not OR - #48

dhcole:
To address inconsistencies in search, documented in #947, this rewrites most of the search views. There used to be a lot of unused or unnecessary code for server-side search, even-though all tasks are sent to the client on page load. 

This adds in a client-side filter function that will filter results based on a search term string or a filter object, e.x. `{ state: ["open", "assigned"] }`. These values can also be read from query parameters `search` and `filter`, which is a JSON representation of the filter object.

This also simplifies the search bar so that it automatically limits tasks based on entered search terms.

There are a few loose ends to finish up, so this is still a work in progress for now:

- [ ] ~~infinite scroll doesn't respect the applied filter and loads additional tasks~~
- [x] need to add in UI for additional filters
- [x] fix profile list (which is a different view)

Resolves #947


dhcole:
Updated to get search working on the profile list. Search now applies to both the map and the list, so you can do things like this:

screen shot 2015-09-24 at 5 10 23 pm

Refs #830

Just need to add in links on location tags to close that out.

dhcole:
Adds in #978, @austinkeeley's work on a Leaflet map. There's some bloat with both D3 and Leaflet, but we can optimize that later. The issue with #978 was in fact missing CSS.

dhcole:
I think this is ready for review. The infinite scroll issue is separate and a pre-existing bug. We can try to fix that up later and turn it on to improve rendering time for these lists. It's not unbearable yet without it.

assets/js/backbone/apps/browse/views/browse_main_view.js
```diff
@@ -17,248 +17,125 @@ var popovers = new Popovers();
...
-        if ( e.choice.hasOwnProperty("unmatched") && e.choice.unmatched ){
-          //remove the hint before adding it to the list
-          e.choice.name = e.val;
+
+      if (!fullTimeTag) return item;
+      if (fullTimeTag && userAgency && (timeRequiredTag.data.agency.id === userAgency.id)) return item;
+    }).filter(function(data) {
+      var searchBody = JSON.stringify(_.values(data)).toLowerCase();
+      return !term || searchBody.indexOf(term.toLowerCase()) >= 0;
+    }).filter(function(data) {
```
jeremiak: This `.filter` chain is pretty nifty, nice work!

jeremiak:
This looks great to me, and its pretty quick! :+1:

PR #1016, opened 2015-09-24
Use tasks in 'open' state to calculate large task count

### LINKED ISSUE #1014: Verify the task count on the dashboard
I see "There are 85 open opportunities right now." on the dashboard but only 10 on /tasks and in the admin section. Possibly a state filter missing?

jeremiak:
The task/opportunity count on the dashboard can show a number much higher than the number of open tasks. Previously we were using any task that wasn't in the 'completed' state, but now we are ONLY using tasks that are in the 'open' state.

To fix #1014


dhcole:
:+1:

PR #1025, opened 2015-09-30
Prompt users to fill out their profiles if they haven't already

### LINKED ISSUE #463: Prompt people to fill out their profile, if they haven't already, when they volunteer
We need to design an appropriate series of prompts or something that leads to them editing their profile

jeremiak:
If a user does not have their agency or location filled out when they click "Participate" on an opportunity, we want to prompt them to do so. This PR changes behavior so that when such a user wants to participate, we present them with a modal to request their agency and location so that we can update their profile with that information.

@dhcole finally ready for your review

To close #463


dhcole:
:+1:

PR #1044, opened 2015-10-05
Display select for person count on new opportunity view

### LINKED ISSUE #1003: As a task creator, I want to choose how many people I need to participate when creating my opportunity.
Include drop down of number of people needed to participate in the opportunity. Options should be the same as previously existed: 1 person, 2-5 people, More than 5 people, A Team
[18F_MIDAS_Taskv2.3.pdf](https://github.com/18F/midas/files/2344/18F_MIDAS_Taskv2.3.pdf)

jeremiak:
This enables new opportunity creators to specify how many participants they are looking for and fixes #1003 

@dhcole and @GailSwanson - this is ready for review

### DIFF
--- a/assets/js/backbone/apps/tasks/new/views/task_form_view.js
+++ b/assets/js/backbone/apps/tasks/new/views/task_form_view.js
@@ -34,7 +34,13 @@ var TaskFormView = Backbone.View.extend({
 
   initializeSelect2Data: function () {
     var self = this;
-    var types = ["task-time-required", "task-length", "task-time-estimate", "task-skills-required"];
+    var types = [
+      "task-time-required",
+      "task-length",
+      "task-time-estimate",
+      "task-skills-required",
+      "task-people"
+    ];
 
     this.tagSources = {};
 
@@ -134,6 +140,11 @@ var TaskFormView = Backbone.View.extend({
       width: 'resolve'
     });
 
+    self.$("#people").select2({
+      placeholder: 'People required',
+      width: 'resolve'
+    });
+
   },
 
   initializeTextArea: function () {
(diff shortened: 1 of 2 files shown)


assets/js/backbone/apps/tasks/new/views/task_form_view.js
```diff
@@ -231,13 +243,15 @@ var TaskFormView = Backbone.View.extend({
...
-        tagLocation   = this.$("#task_tag_location").select2('data');
+        tagLocation   = this.$("#task_tag_location").select2('data'),
+        peopleCount   = this.$("#people").select2('data');
 
     // check for the presence of data in these fields
     // no data means no tags are supplied
     // don't send those tags because the API returns 500
     if (tagSkills != []) tags.push.apply(tags, tagSkills);
     if (tagLocation != []) tags.push.apply(tags, tagLocation);
+    if (peopleCount != []) tags.push(peopleCount);
```
dhcole: If I'm following this correctly, `peopleCount` would be an array, so pushing this onto the end of tags would be adding the array, instead of just the object. That's why the lines above use apply as a mutator form of concat.

dhcole:
Just one note above; otherwise, this looks good.

assets/js/backbone/apps/tasks/new/views/task_form_view.js
jeremiak: I don't think this is true, because the field that `peopleCount` pulls from is not a multi-select like skills and location, but will always just have one value which is an object and so it can just be pushed onto the `tags` array.

dhcole: I missed the multi-select vs single-select difference. Thanks for explaining that.

PR #1045, opened 2015-10-05
Show all participants of a task in dashboard feed

### LINKED ISSUE #1034: As logged in user, I need The Achievements Wall to reflect who completed tasks with their photo.
There seems to be a problem between the with the Achievements Wall and the the profiles.  I noticed that 2 of the people that worked on the task Building the White House led Citizen Science Toolkit were not on the Achievements wall and were not seen on the task Those people are Laura Kahn and Hutch Brown.  I sent them and email asked them to press the I'm in button again so they would show up with the other three that were there. (This is what it looked like)
image

So Laura Kahn clicked the I'm in button and on the achievements wall Bernetta's photo disappeared and their is no photo of Laura or circle where she would go.

image

jeremiak:
To fix #1034 we want to show all the participants for a given task. Previously we were explicitly selecting for just the first three, but now that has been altered to just display all participants for a given task.

### DIFF
--- a/assets/js/backbone/apps/home/templates/home_badges_feed_template.html
+++ b/assets/js/backbone/apps/home/templates/home_badges_feed_template.html
@@ -20,8 +20,7 @@
           </div>
 
           <div class="activity-feed--avatars">
-            <% window.x = b.participants; %>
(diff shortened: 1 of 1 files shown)


jeremiak:
@dhcole ready for your review

PR #1048, opened 2015-10-07
Refactor activity endpoints to be faster

dhcole:
I noticed the API endpoints on the user dashboard were taking a few seconds. It turns out these were pretty easy to optimize, so here it is.


jeremiak:
:fast_forward: :+1:

jeremiak:
Sorry @dhcole I thought I Had merged this after 78ff94f but apparently not. It all looks good to me, so I'll get it merged.

PR #1049, opened 2015-10-07
Refactor duplicated function expressions for enabling and disabling a…

kencheung:
… user


dhcole:
Thanks for submitting this @kencheung. We'll review it for our next release.

api/controllers/UserController.js
```diff
@@ -258,16 +258,20 @@ module.exports = {
     });
   },
 
-  // Enable a disabled user
-  enable: function (req, res) {
-    // policies will ensure only admins can run this function
-    User.findOneById(req.route.params.id, function (err, user) {
+  findUserById: function(res, err, user, userDisabled, saveErrMsg) {
```
jeremiak: Thanks for submitting this PR, refactoring is always appreciated :smile: 

In this case, I'm not sure I'm sold on introducing a function that takes five positional arguments. I find it hard to remember a function who's signature includes five arguments and know what the order is. Instead, how about submitting some of these as keys on an object that is passed in. Something similar to what you see is the function signature for `res.send()` on line 265.

By way of example, `findUserById` might take something like `(res, err, { user: user, userDisabled: false, errMsg: ''})`.

Thoughts?

kencheung: That improves readability. I have revised the PR as suggested. See if it looks better to you.

jeremiak:
Hi @kencheung - thanks for this and for making some of the changes I suggested.

Can you get this branch to pass tests? You should be able to run them locally with `npm run test` and have it exit successfully. As you can see above, Travis is reporting that this branch fails the tests.

kencheung:
You are welcome. I have made the branch to pass tests as suggested.

dhcole:
Thanks @kencheung! :ship:

PR #1050, opened 2015-10-09
Use sudo for Travis

dhcole:
Travis deployments require `sudo`.


jeremiak:
:+1: I like that this means we're getting automatic deployments :smile:

PR #1057, opened 2015-10-13
Update CF endpoint

dhcole:
We switched endpoints for the CF endpoint. https://cloudgov.statuspage.io/incidents/fm8ksv9byzdb

@jeremiak one more to get CD working.


jeremiak:
:+1:

Continuous deployment, here we come!

PR #1063, opened 2015-10-15
An admin can add participants to an opportunity

### LINKED ISSUE #1036: As an admin, I would like to be able to associate participants with a task
this should not generate an email, it is primarily for making historical corrections (though I imagine the need may arise periodically in the future)

jeremiak:
This PR should close #1036, and adds a dropdown menu for admin users to the opportunity edit view. The admin user can select any user in the database to add as a participant and select to generate a notification as well.

Once the admin user completes other edits and saves their changes, the participant relationship will be created.

I think this is ready for review @dhcole.

### DIFF
--- a/api/controllers/VolunteerController.js
+++ b/api/controllers/VolunteerController.js
@@ -9,6 +9,7 @@ module.exports = {
 
   create: function (req, res) {
     var v = _.extend(req.body || {}, req.params);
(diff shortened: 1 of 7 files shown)


jeremiak:
Ok @dhcole, finally ready for you.

dhcole:
Nice work @jeremiak. :ship:

PR #1071, opened 2015-10-16
Skip participant records for missing tasks or users

dhcole:
Quick fix for an error that prevented the participants table from loading.


jeremiak:
Looks good to me!

PR #1081, opened 2015-10-20
Display working groups by default in UI

jeremiak:
In commit 90d08bc8a59e020e1b4cfe488686c0a937b18bab the setting default was changed from displaying working groups to hiding them. This reverts that back so that working groups are displayed by
default.

cc @ultrasaurus


dhcole:
:+1: Tests need to run again after merging in the latest from `dev`; then ready to :ship:

Issue #1043, opened 2015-10-05
"Agency" is not configurable...


ultrasaurus:
I thought @brandtheisey had this working, but from looking at the code, it appears that the word "Agency" isn't configurable

ultrasaurus:
@mokk083 talked to Rhea today who said you wanted to change "Agency" to "Division" -- it appears in a couple of placed in the UI.  The right fix would be needed is to make that text modifyable, adding to `locales/en/translation.json` and then referring to it via i18n library in the HTML

ultrasaurus:
for a example of how we do configurable text, check out:  https://github.com/18F/midas/blob/devel/assets/js/backbone/apps/tasks/new/templates/task_form_template.html#L3

PR #1089, opened 2015-10-22
Agency not configurable

awhig:
Made the term "Agency" a translatable term. The term is used in several places
in the UI. It is also a "tag" item, so certain parts of the UI already were
translatable by changing the name and plural properties in
assets/js/backbone/config/tag.js.

Fix for Issue #1043


rogeruiz:
Hey @awhig. Thanks for the PR.

PR #1091, opened 2015-10-25
Update to Layout in Install Docs and Addition of Table of Contents

gilchenzion:
and I moved the Windows instructions up to where the others Platforms were. When I was going through the install process on the Mac, I skipped over a couple steps since they seemed like they were located under the Linux install setups. This hopefully clarifies that and adds a table of contents to the top of the install for better navigation.


rogeruiz:
Thanks @gilchenzion. The documentation is more linear now and easier to navigate and deep-link to particular topics.

Issue #636, opened 2015-03-09
when I click on a link in opportunity description, it should open in a new window or tab


ultrasaurus:
On an opportunity detail page:
![image](https://cloud.githubusercontent.com/assets/41906/6561232/ca4ac3ce-c64c-11e4-9085-2226a9ed437a.png)

I expected the link to go to a new tab or window

Sujakosa:
Hello, Ultrasaurus. Try making the target of the link blank. A blank-target usually opens a new window. I hope that helps.

ultrasaurus:
@Sujakosa yep.  I know how to do it -- just need time to hunt down where it is in the source code and make the change.  I thought this might be a fun thing for someone new to the project!

awhig:
Markdown doesn't let you put a target attribute on the link.

One option is to use some jQuery to set a blank target on any links on the page that link out of the site. Would that work?

$(document.links).filter(function() {
    return this.hostname != window.location.hostname;
}).attr('target', '_blank');

If that's too wide, then maybe only do that to links that are inside of the description div.

<div class="task-show-description">
</div>

http://stackoverflow.com/questions/4425198/markdown-target-blank

awhig:
@dhcole Would the above solution work for you?

Rich

PR #1092, opened 2015-10-25
Now rendering external links with target=_blank

techthumb:
I'll be opening a PR on marked for this change, with an option to externalize links.

However, in the meanwhile, this should address #636


jeremiak:
Hey @techthumb thanks for the PR.

Just a few quick thoughts. It generally isn't a good idea to change a library's prototype somewhere in the code without being super explicit about it. What I might change is to have the util file export the link rendering function and then explicitly set that as the renderer in the options in `app.js`.

Does that make sense?

techthumb:
Yes! Fixing now!

PR #1093, opened 2015-10-25
Make the good/bad email icon not appear as a clickable link

jsfeb26:
On the signup page I changed the good/bad email icon so that it doesn't appear clickable.  I changed the anchor tag to a span so that it doesn't show a dead link at the bottom.  I then added a new class so that the cursor isn't a pointer and doesn't look like a link.


ultrasaurus:
this looks good to me -- seems like it would be helpful for usability!  checked out the code, which seems like a reasonable approach.

PR #1100, opened 2015-10-26
Ensure that user activity view on profile is populated

### LINKED ISSUE #1098: I don't see tasks where I am a participant on my profile...
https://openopps-staging.18f.gov/profile/3

I tried changing task state from open to assigned, even to completed and I never saw anything on my profile....
image

Here's the task where I'm participating:  https://openopps-staging.18f.gov/tasks/420

jeremiak:
To fix #1098, we want to only pass valid task objects to the template. This PR changes some front end code to filter out any task object that doesn't have the `createdAt` field.

### DIFF
--- a/assets/js/backbone/apps/profiles/show/views/profile_activity_view.js
+++ b/assets/js/backbone/apps/profiles/show/views/profile_activity_view.js
@@ -18,9 +18,13 @@ var ProfileActivityView = Backbone.View.extend({
 
   render: function () {
     // sort initially by date, descending.
-    var results = this.options.data.sort(function (a, b) {
-      return new Date(b.createdAt) - new Date(a.createdAt);
-    });
+    var results = this.options.data
+        .filter(function(i) {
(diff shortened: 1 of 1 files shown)


dhcole:
So do you think the issue was that an old task was missing `createdAt`, causing the JS execution to crash, and stopping loading of more recent tasks? It's hard to think of how a task wouldn't have a createdAt date since that's set automatically by the ORM (unless it was accidentally cleared) but I could see something like that happening in the past and affecting old data.

jeremiak:
It wasn't that the task didn't have a `createdAt` field, it was that the first object in the JSON array was `null` followed by four valid task objects. The filter just removes anything that doesn't have a `createdAt` field, null included.

dhcole:
:+1: I understand now. This looks like a good front-end solution. The API shouldn't return `null` objects and probably could be fixed directly, but this is a good way to go to catch this or any other error that would crash the client js.

PR #1102, opened 2015-10-27
Hide "I'm in" if opportunity is closed

### LINKED ISSUE #984: I should not be able to click "I'm In" on a closed opportunity
The button should be disabled or maybe just hidden.  I think that it used to behave that way and must have broken at some point

jeremiak:
To close #984, the “I’m In” button will be hidden if a task is in either the "completed" or the "archived" state.


dhcole:
Actually, does it make sense to flip this around and only show the i'm in button for `model.state === 'open'`? I'm thinking I'm In doesn't make sense for draft of assigned tasks either.

jeremiak:
Great point @dhcole. Change made and pushed.

PR #1109, opened 2015-10-29
Production push of 0.7.6 - try two

jeremiak:
The last production push deploy failed because of a missing env variable on Cloud Foundry. This is the same release, with a minor change to `.travis.yml`.


jeremiak:
Reverting since production appears to be broken by this deploy

PR #1110, opened 2015-10-30
Revert "Production push of 0.7.6 - try two"

jeremiak:
Reverts 18F/midas#1109


jeremiak:
0.7.6 actually included a database migration (and should've been called 0.8.0 from the start), but the migration didn't run which is causing the production application to fail.

PR #1111, opened 2015-10-30
Include instructions for running a migration

jeremiak:
Our deploy documentation was missing key instructions on how to run a database migration. I've tried to add it here.

@dhcole anything you would change/add?


DEPLOY.md
```diff
@@ -25,6 +25,18 @@ Once a release has been validated on staging, open a new PR for the release agai
...
+### Deploying to Cloud Foundry with a migration
+
+If you have a migration that needs to be run, you'll need a few things:
+
+0. `cf-ssh` installed on your machine
+1. "Access a postgres database using cf-ssh" which can be found on the [cloud.gov documentation](https://docs.cloud.gov/apps/databases/)
+
+Using `cf-ssh`, access the production instance of Open Opportunities. Make sure you're on the proper branch on your local environment. This is because `cf-ssh` uses the local files to build the new instance. Please note that this might take a few minutes. Once you're into the production box follow the instructions on the cloud.gov documentation for installing `psql`. We need `psql` because many of the migration scripts we have in `/tools/postgres/migrate` are bash scripts that run queries through `psql`.
+
+Now you can simply run `./tools/postgres/migrate/yourscript.sh` and the migration should be in place.
```
dhcole: This is a bit more complicated because the .sh files usually hardcode the database and user. Instead you need to edit this file to use `$DATABASE_URL` or copy the `-c` command and run it from the command line according to the instructions in https://docs.cloud.gov/apps/databases/

jeremiak: I've gone ahead and updated all of the old migration scripts to use `$DATABASE_URL` (and added it as an environment variable in my local `.env`) instead of the hardcoded DB name. Seems nicer to be able to just run the same file and change an env variable instead of copy and paste things out of a file.

ultrasaurus:
@jeremiak this looks like it would affect the  `npm migrate` command that is part of the initial dev setup.  Maybe `tools/postgres/init.sh` should set `DATABASE_URL=midas` if not defined, otherwise I think we would always have to require setting that environment variable

jeremiak:
Another issue that should be addressed before this is merged:

- [ ] We still have the table owner hardcoded ([i.e `5.sh`](https://github.com/18F/midas/blob/dev/tools/postgres/migrate/5.sh#L12)) as `midas`, yet that is set on as an environment variable on Cloud Foundry. See

dhcole:
> We still have the table owner hardcoded (i.e 5.sh) as midas, yet that is set on as an environment variable on Cloud Foundry.

Right. I'm not sure off hand of an easy was to make this work with CF, since the CF psql service sets a dynamic user name for the DB.

I do want to take a look in automating this in #1126 but haven't had a chance to do it this week.

I'd say it's worth merging this as is since it improves documentation around an undocumented part of the process, recognizing that there's still more work to do to get it :100:

tools/postgres/migrate/1.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Add schema table
-psql -U midas -d midas -c "CREATE TABLE \"schema\" (
+psql -U midas -d $DATABASE_URL -c "CREATE TABLE \"schema\" (
```
dhcole: In this and subsequent occurrences, `$DATABASE_URL` is a full connection string that includes a user, password, and database name, so the command should actually start with `psql $DATABASE_URL` instead of ~~`psql -U midas -d $DATABASE_URL`~~

rogeruiz: :+1:

rogeruiz:
I'll be merging this into `dev` and make it part of the current `0.8.0` release since it's so necessary for DB migrations.

Issue #906, opened 2015-07-13
Application CSS & Theme CSS refactor; Theme / Platform synchronization / overrides


jeremiak:
Currently, we only have only one CSS file that is used for the application instance. For Open Opportunities this means that we have a specific theme CSS that replaces the midas CSS. This introduces yet another place that we have to make CSS changes and has actually resulted in a few bugs.

We would prefer to have the CSS set up in such a way that the Midas CSS is used and any overriding of defaults occurs in the theme's CSS.

## List taken from @dhcole
- [x] Most of [this repo](https://github.com/18F/open-opportunities-theme) can be merged into the main application
- [x] Things like logos, home page image, translation file would need to be separate
- [x] Having those files extend / cascade default application files would avoid past issues with deployment
- [x] Currently files are rsync'd over the existing application files so everything needs to be duplicated in both files

dhcole:
Capturing some ideas from standup:

- Most of this repo can be merged into the main application
- Things like logos, home page image, translation file would need to be separate
- Having those files extend / cascade default application files would avoid past issues with deployment
- Currently files are rsync'd over the existing application files so everything needs to be duplicated in both files

rogeruiz:
I'm proposing that [this](https://github.com/18F/midas/issues/887) other ticket get merged or at least cross-referenced into the same pull request for this issue since they're related. Having a pre-processor, or "file funnel" as I like to call it, run through the current files and concat them can really lead the way to setting up an "*extend / cascade default app files*" much easier than refactoring the files first.

ultrasaurus:
@rogeruiz feel free to add detail to the description of this ticket and close the other as a dup of this one.

rogeruiz:
Keeping the discussion going with #887, to allow easier "defaults" and theming of the application I will be migrating the stylesheets to use Sass and Bourbon. The goal of this isn't to re-write all the styles for the project but rather have the original CSS files broken down into partials and also keep as much of the original CSS there run through the Sass preprocessor in order to best move forward making the stylesheets more modular.

PR #1121, opened 2015-11-06
Silence badge notifications and alter historical badge assignment script

jeremiak:
This PR addresses issues that I ran into getting the badges feature onto staging:

0. Silences badge notifications as a default. Since we don't currently have a way to pass that option all the way through the historical assignment would've triggered hundreds of notifications that should've happened in the past. 
0. Updates the `assignHistoricalBadges.js` script to use the updated badge API and to provide feedback to the user running the script.
0. Fix an undefined variable error in `User.js`
0. Remove a client side logging statement that snuck through

Just as a note, I did run into the perennial issue with the `open-opportunities-theme` repo, namely I forgot to update it. I wanted to document that #906 would have been useful in this case.

As a result of the way Cloud Foundry works, when I ran `cf-ssh` with these changes in the working directory it uploaded them to the `ssh` host. Consequently, I was able to get this code on staging and run the updated historical assign script. I did this because we have an interest in getting badges on staging for socialization, though this PR should be merged and then redeployed to staging so that the `staging` branch reflects what is on the environment.

@ultrasaurus would you mind reviewing? I can then PR against `staging`.

### DIFF
--- a/api/models/Badge.js
+++ b/api/models/Badge.js
@@ -87,23 +87,28 @@ module.exports = {
    *
    * @param {object} task
    * @param {object} user
+   * @param {object} opts - optional
    * @param {function} done
    */
-  awardForTaskCompletion: function(task, user, done) {
+  awardForTaskCompletion: function(task, user, opts, done) {
     var completedAwards = {
           1: 'newcomer',
           3: 'maker',
           5: 'game changer',
           10: 'disruptor',
           15: 'partner'
-        };
+        },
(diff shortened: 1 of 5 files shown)


ultrasaurus:
I see that it is silencing badge notifications by default.  So, how will notifications happen for new badges?  am I missing something?

jeremiak:
Good point @ultrasaurus. Updated the code to silence only the historical badge assignments.

ultrasaurus:
looks good :+1:

PR #1122, opened 2015-11-06
Historical badges are silent

jeremiak:
We want historical badges to be silent, but all future badge creation to create notifications. This adjusts the data model and one time script to reflect that.


ultrasaurus:
@jeremiak confused on process here... going from dev to staging would have already been reviewed.  Any reason for this PR to be open?

PR #1139, opened 2015-11-14
Admin user metrics extraction

baccigalupi:
I looked at the CodeClimate metrics and saw that the AdminController was the source of the highest technical debt with a combination of high churn and high complexity. This set of commits extracts the decoration of users with metrics into a separate service module. That service module houses a class that is unit tested.

Also, in the course of unit testing, I found a bug in the reporting of created projects: the wrong attribute was being used to look up the association.

There is more work that can be done to make the queries use associations, and do queries in a more uniform way, but this set of commits will hopefully make it easier for the next person if more work needs to be done in this area.

### DIFF
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -9,6 +9,8 @@ var json2csv = require('json2csv');
  */
 var async = require('async');
 
+var addUserMetrics = require(__dirname + '/../services/utils/userMetrics').add;
+
 module.exports = {
 
   /**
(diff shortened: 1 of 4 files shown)


api/controllers/AdminController.js
```diff
@@ -33,150 +35,29 @@ module.exports = {
...
-            q: query
+      .where(where)
+      .sort({ createdAt: 'desc' })
+      .paginate({ page: page, limit: limit})
+      .exec(function (err, users) {
+        if (err) { return res.send(400, { message: 'Error looking up users', err: err}); }
+        // count the total number of users
+        User.count(function (err, count) {
+          if (err) { return res.send(400, { message: 'Error counting users', err: err}); }
+          async.each(users, addUserMetrics, function(err) {
```
baccigalupi: This set of changes is really just reformatting to make the code clearer via indentation. No changes were made in this section to the actual code.

api/models/Task.js
```diff
@@ -101,7 +101,7 @@ module.exports = {
         case 'completed':
           values.completedAt = new Date();
           action = 'task.update.completed';
-          task.volunteersCompleted();
+          task && task.volunteersCompleted();
           break;
```
baccigalupi: I am not sure if this is a bug, or intended. I wasn't able to create a task in the completed state because while the task data exists in this before hook, the task object with this method does not. It hasn't been saved yet and therefore can't be found in the database query. I had to add this guard to the before save hook in order to create a task in this state.

api/services/utils/userMetrics.js
```diff
@@ -0,0 +1,135 @@
+var async = require('async');
+var _ = require('underscore');
+
+function Decorator(data) {
+  this.data = data;
+};
+
```
baccigalupi: I don't know if the class extension of a prototype is an acceptable way of working with classes in this project. I did it as a last refining step when I wanted to compare methods. Having the methods in a hash made it easier for me to see similarities and differences between methods. Earlier commits build functions on the prototype. I left those commits currently unsquashed for comparison.

dhcole:
Thanks @baccigalupi. This looks great. I'll do a review and we'll get it ready for our next release. I appreciate the contritbution!

dhcole:
Sorry for the delay. This is great. We haven't used a decorator pattern on the application yet, but it looks like an interesting approach we should keep in mind for other areas we refactor. The expanded test coverage (and bug fix) is really helpful too!

Looks like consistent response data and performance improvements too:

image

Thanks @baccigalupi! :ship:

baccigalupi:
Just circling back on this. Glad it helped!

PR #1165, opened 2015-12-01
Creator badges

rogeruiz:


api/models/User.js
```diff
@@ -57,6 +57,12 @@ module.exports = {
       defaultsTo: 0
     },
 
+    // Store the number of assignedTasks
+    //assignedTasks: {
+      //type: 'INTEGER',
+      //defaultsTo: 0,
+    //},
+
```
ultrasaurus: should we remove this commented out code before checking in?

api/models/User.js
```diff
@@ -68,26 +74,38 @@ module.exports = {
...
+      user.completedTasks = 1 + user.completedTasks;
+
+      this.save( function ( err, u ) {
+
+        if ( err ) { return sails.log.error( err ); }
+
+        Badge.awardForTaskCompletion( task, user, opts );
+
+      } );
+
```
ultrasaurus: a little nit picky, but `+= 1;` is more readable than `user.completedTasks = 1 + user.completedTasks;`
I think these two code blocks are actually identical

api/models/User.js
rogeruiz: Ah yeah, I'll update it.

api/models/User.js
rogeruiz: Sounds good, I'll revert this as well.

ultrasaurus:
did @jeremiak have the chance to review this?

ultrasaurus:
by the way, I did review all the code, and *think* I understand it and it all looks good to me!  So, we don't need to wait for @jeremiak but would be nice if he's avail

PR #1178, opened 2015-12-04
Discrepant UI edit

### LINKED ISSUE #1170: UI discrepancy when editing profile
This bug is on the UI layer only as Backbone isn't checking permissions before presenting the user with the edit UI. The API layer is protected and displays the appropriate message if the user tries to click "Save Profile".

**Steps to reproduce:**

1. Login with your Open Opportunities Account
2. In the upper right hand corner click on your profile name
3. Click Profile
4. Click Edit Profile
5. Change the number from your profile number to another profile number in the URL `/profile/2 => /profile/5`

rogeruiz:
This PR should fix #1170 when it's merged in.

### DIFF
--- a/assets/js/backbone/apps/profiles/show/views/profile_show_view.js
+++ b/assets/js/backbone/apps/profiles/show/views/profile_show_view.js
@@ -32,14 +32,30 @@ var ProfileShowView = Backbone.View.extend({
 
   initialize: function (options) {
     var self = this;
+    var model = this.model.toJSON();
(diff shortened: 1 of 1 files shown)


rogeruiz:
I've decided against visiting integration testing on this using CasperJS until #566 is figured out a bit more. The CasperJS test suite with was giving me some issues whenever I was trying to navigate to another page without interacting with the application's links.

```
casper.thenOpen( server.env.TEST_ROOT + '/profile/2/edit', function () { /* assertion * / } );
```

Issue #1175, opened 2015-12-03
disabling users seems to be broken


ultrasaurus:
spins forever and I see 502 error in Javascript console

![image](https://cloud.githubusercontent.com/assets/41906/11570848/425cf1fe-99ae-11e5-9d60-46aa5847c419.png)

Issue #1172, opened 2015-12-02
travis should run `npm run demo`


ultrasaurus:
We want to make sure the new developer experience is always good (and also right now I think the demo script does some integration testing that we don't have covered elsewhere)

add a line after `npm install` [here](https://github.com/18F/openopps-platform/blob/dev/.travis.yml#L5-L6) for the shell command `npm run demo`

or to be more specific, you could add a `before_script` step to the file

rogeruiz:
Running into issues with this because of the way `npm run demo` requires a connection to the DB for running it's test. Seems like the demo command may need to be refactored to avoid the DB calls to successfully run in Travis.

```
$ npm run demo
> openopps-platform@0.8.2 demo /home/travis/build/18F/openopps-platform
> mocha test/test.upstart.js --development --recursive test/demo
lifting sails: env=development
Loading...  /home/travis/build/18F/openopps-platform/config/application.js
Loading...  /home/travis/build/18F/openopps-platform/config/assets.js
Loading...  /home/travis/build/18F/openopps-platform/config/bootstrap.js
Loading...  /home/travis/build/18F/openopps-platform/config/connections.js
Loading...  /home/travis/build/18F/openopps-platform/config/controllers.js
Loading...  /home/travis/build/18F/openopps-platform/config/csrf.js
Loading...  /home/travis/build/18F/openopps-platform/config/email.js
Loading...  /home/travis/build/18F/openopps-platform/config/globals.js
Loading...  /home/travis/build/18F/openopps-platform/config/http.js
Loading...  /home/travis/build/18F/openopps-platform/config/io.js
Loading...  /home/travis/build/18F/openopps-platform/config/models.js
Loading...  /home/travis/build/18F/openopps-platform/config/policies.js
Loading...  /home/travis/build/18F/openopps-platform/config/routes.js
Loading...  /home/travis/build/18F/openopps-platform/config/session.js
Loading...  /home/travis/build/18F/openopps-platform/config/env/development.js
Loading...  /home/travis/build/18F/openopps-platform/config/settings/analytics.js
Loading...  /home/travis/build/18F/openopps-platform/config/settings/auth.js
Loading...  /home/travis/build/18F/openopps-platform/config/settings/sources.js
Loading...  /home/travis/build/18F/openopps-platform/config/settings/tags.js
Loading...  /home/travis/build/18F/openopps-platform/config/sockets.js
Loading...  /home/travis/build/18F/openopps-platform/config/version.js
[...]

Issue #1171, opened 2015-12-02
error turning a model into an object...


ultrasaurus:
this might be some issue with my local install, but tracking it here since others might run into it later

```
$  git clone git@github.com:18F/openopps-platform.git
$  cd openopps-platform/
$  npm install
$  npm run migrate
$  npm run init
$ npm run demo

> openopps-platform@0.8.0-rc4 demo /Users/sarahallen/src/18F/openopps-platform
> mocha test/test.upstart.js --development --recursive test/demo



lifting sails: env=development
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/application.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/assets.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/bootstrap.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/connections.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/controllers.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/csrf.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/email.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/globals.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/http.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/io.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/models.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/policies.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/routes.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/session.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/env/development.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/settings/analytics.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/settings/auth.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/settings/sources.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/settings/tags.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/sockets.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/version.js
Loading...  /Users/sarahallen/src/18F/openopps-platform/config/views.js
[...]

PR #1187, opened 2015-12-10
Revert `UserController` enable and disable

rogeruiz:
Because the `user` object needs to be looked up and modified before being saved, leaving the code as it was before makes the most sense without refactoring this out to a `UserUtil` service.

This PR should touch multiple issues, because of how related they are to the `User` model.

- [x] **#1175** *disabling users seems to be broken*
- [x] **#1171** *error turning a model into an object...*
- [ ] ~~**#1172** *travis should run `npm run demo`*~~

I'll have to verify the changes made here run successfully with `npm run demo`. The previous errors I was experiencing were related to these two methods `enable` and `disable`.

### DIFF
--- a/api/controllers/UserController.js
+++ b/api/controllers/UserController.js
@@ -257,22 +257,16 @@ module.exports = {
     });
   },
 
-  findUserById: function(res, err, userInfo) {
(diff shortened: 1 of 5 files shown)


rogeruiz:
Running into issues with this because of the way `npm run demo` requires a connection to the DB for running it's test. Seems like the demo command may need to be refactored to avoid the DB calls to successfully run in Travis.

PR #1196, opened 2015-12-15
refactor so “impact” can be one line

ultrasaurus:
need to go with current dashboard feature set
markdown (including line breaks!) is a future feature

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -8,7 +8,11 @@ description: >
   Open Opportunities is creating a network of federal employees that
   will lead to a more effective, efficient, responsive government by sharing skills
   and collaborating on projects that support the mission of an agency or the
-  government as a whole.
+  government as a whole.    Only 37% of federal employees surveyed believe that
+  creativity & innovation are rewarded. From a total audience of 2.7 million
+  federal employees, the initial target audience are motivate, mission-driven
+  individuals who identify and solve problems for our nation.
+
 
 partners:
 - General Services Administration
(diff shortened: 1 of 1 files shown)


ultrasaurus:
@rogeruiz can you proof-read and merge if it looks good?

ultrasaurus:
by the way, here's a preview using new dashboard feature that is not yet merged (partner section is missing cuz preview tool doesn't support it)

image

.about.yml
```diff
@@ -8,7 +8,11 @@ description: >
   Open Opportunities is creating a network of federal employees that
   will lead to a more effective, efficient, responsive government by sharing skills
   and collaborating on projects that support the mission of an agency or the
-  government as a whole.
+  government as a whole.    Only 37% of federal employees surveyed believe that
```
rogeruiz: I'm going to clean up this whitespace between `whole.` and `Only`.

.about.yml
```diff
@@ -18,7 +22,7 @@ partners:
 stage: live
 
 impact: >
-  65 agencies, 3000 subscribers\n\n278 projects completed, 40 in progress\n\ntotal market: 2.7 million federal employees\n\noutcome: skilled, efficient, agile workforce
+  Amplify the work of 1 Million federal employees.
```
rogeruiz: I'm also going to update down-case `Million` here as well.

rogeruiz:
Once these builds pass, I'll merge this in and include it in `0.8.2`

PR #1213, opened 2016-01-09
Refactor: extracts duplication from LikesController

baccigalupi:
After the last batch of refactors the LikesController showed up
        as having the most technical debt. This seemed mostly related to
        duplicate methods.

        There is still some duplication in the controller actions that I
        haven't worked with, but the like and unlike related actions
        have be cleaned up:

        * Extracts a LikeActions module with three 'classes'
                * top level handler that takes dependencies and has two
                  methods: 'like', 'unlike'
                * classes for liking and unliking

### DIFF
--- a/api/controllers/LikeController.js
+++ b/api/controllers/LikeController.js
@@ -5,26 +5,7 @@
  * @description	:: Contains logic for handling requests.
  */
 
-var like = function (l, cb) {
-  Like.findOne({ where: l }, function (err, existing) {
-    if (err) { return cb(err, null); }
-    if (existing) { return cb(null, existing); }
-    Like.create(l, function (err, newLike) {
-      if (err) { return cb(err, null); }
-      return cb(null, newLike);
-    });
-  });
-};
-
-var unlike = function (l, cb) {
-  Like.findOne({ where: l}, function (err, newLike) {
-    if (err) { return cb(err); }
-    if (!newLike) { return cb(null); }
-    newLike.destroy(function(err) {
-      return cb(err);
-    });
-  });
-};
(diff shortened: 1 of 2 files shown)


rogeruiz:
:+1: looks good @baccigalupi Thanks for creating this service for Likes/t/u. I can merge this in for this deployment. I'd like to update the comments around [two](https://github.com/baccigalupi/midas/blob/like-controller-cleanup-v1/api/controllers/LikeController.js#L122) of the [methods](https://github.com/baccigalupi/midas/blob/like-controller-cleanup-v1/api/controllers/LikeController.js#L98) you refactored to reflect the actual names in the route. I'll tag them in "Files Changed" view as well. Thanks again.

api/controllers/LikeController.js
```diff
@@ -157,11 +122,7 @@ module.exports = {
    * Syntax: Call /like/unlikeu/:userId where :id is the userId
```
rogeruiz: Should read

```
 * Syntax: Call /like/unliket/:userId where :id is the userId
```

api/controllers/LikeController.js
```diff
@@ -121,11 +98,7 @@ module.exports = {
    * Syntax: /like/likeu/:userId where :id is the userId
```
rogeruiz: Should read

```
 * Syntax: Call /like/liket/:userId where :id is the userId
```

rogeruiz:
Since these are small changes around the comments, I'm including these changes into the 0.9.x release today and fixing the comments myself. Thanks @baccigalupi for your contributions to the project! :+1:

Issue #1247, opened 2016-02-23
upgrade casperjs causes tests to fail?


ultrasaurus:
I can reproduce the CI issue locally if I delete `node_modules` and do a fresh `npm install`

```
  1) "before all" hook

  0 passing (1s)

  1 failing

  1)  "before all" hook:
     Uncaught TypeError: 'undefined' is not an object (evaluating 'this.options.pageSettings.javascriptEnabled') (/Users/sarahallen/src/18F/openopps-platform/node_modules/casperjs/modules/casper.js:981)
```

ultrasaurus:
suspect it is one of these

```
< ├── casperjs@1.1.0-beta3
---
> ├── casperjs@1.1.0-beta5
```

```
< ├── mocha-casperjs@0.5.4
---
> ├── mocha-casperjs@0.5.5
```

ultrasaurus:
This seems to be the issue:  https://github.com/nathanboktae/mocha-casperjs/issues/80#issuecomment-183232733  but it should only fails with Phantom 2

ultrasaurus:
this was addressed in 0.10.0 release

PR #1249, opened 2016-02-23
pin versions so that tests pass

ultrasaurus:
seems like newer version of casperjs,
either bet4 or beta5 has some incompatibility

ideally we'd fix the root cause for #1247 but hopefully this will get Travis to be green


ultrasaurus:
whew!  the tests pass, I will merge this

PR #1259, opened 2016-03-17
update to latest schema

ultrasaurus:
also removed reference to jeremiakimelman

### DIFF
--- a/tools/postgres/schema/current.sql
+++ b/tools/postgres/schema/current.sql
@@ -10,14 +10,14 @@ SET check_function_bodies = false;
 SET client_min_messages = warning;
 
 --
--- Name: plpgsql; Type: EXTENSION; Schema: -; Owner: 
+-- Name: plpgsql; Type: EXTENSION; Schema: -; Owner:
 --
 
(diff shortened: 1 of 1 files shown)


ultrasaurus:
@jeremiak I would guess this file was just missed when we did the save-as-draft feature, since we all had existing databases that we upgraded.  I discovered the issue of missing `submittedAt` when I cleaned by db and started fresh, for testing.

jeremiak:
Good catch, @ultrasaurus

tools/postgres/schema/current.sql
```diff
@@ -856,6 +856,7 @@ CREATE TABLE task (
     "updatedAt" timestamp with time zone,
     "deletedAt" timestamp with time zone,
     "publishedAt" timestamp with time zone,
+    "submittedAt" timestamp with time zone,
```
rogeruiz: aha, I see. so does having this column in the DB fix the issue #1260 @ultrasaurus?

ultrasaurus:
@rogeruiz it fixed the bug, but I wasn't able to make a test that failed before this fix... was trying to test drive the fix so if it ever happened again the tests would catch it

PR #1260, opened 2016-03-17
[WIP] submittedAt test should fail

ultrasaurus:
this test does NOT fail locally
I wonder if Sails just assumes that it is ok to have a model with a property
that is only in-memory... I want the Rails equivalent of Task.reload!

submitting as WIP in case someone has any insights on this -- maybe @jeremiak or @rogeruiz ?


ultrasaurus:
decided not to worry about making the test fail when the schema is invalid (since I think that isn't possible now) -- cleaned this up and integrated with:  https://github.com/18F/openopps-platform/pull/1284

Issue #1248, opened 2016-02-23
As a developer, updates to the database should be easier to deploy (migrations in javascript with sails-db-migrate)


ultrasaurus:
We want to use [sails-db-migrate](https://github.com/building5/sails-db-migrate) package so deployments can be executed in Javascript, and developers can easily manage migrations.

We need to support new developers starting with no database, as well as migrations for people at the current version.  (It's ok if people who have several versions ago need to run the current scripts with psql)

`npm run init`,  `npm run install`,  and `npm run demo` should all apply migrations before running

`npm run start` and `npm run watch` should warn with a solution message if there are pending migrations

all tests must pass
documentation must be updated, so as to be clear for new open source contributors or current developer on the team:
- INSTALL.md
- CONTRIBUTING.md

ultrasaurus:
need to test the deployment side of this

PR #1261, opened 2016-03-17
Use db-migrate to Manage Database Migrations

cwarden:
Replace use of SQL scripts run through psql to handle database
migrations with sails-db-migrate.  New migration scripts are written in
javascript.

Create an initial migration script which will load the initial schema.
(It can also be run against existing databases.)

Update the `migrate` npm task to run the new db-migrate migration
scripts.

Update the `init`, `install`, `demo` npm tasks to perform any pending
database migrations.

Update the `start` and `watch` npm tasks to warn the user about pending
migrations.

Update the Travis CI configuration to create the midas database since
`npm install` now runs the database migrations.

*See #1248 and https://micropurchase.18f.gov/auctions/17*

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "build": "$npm_package_config_grunt build",
     "start": "npm run check-migrate && node app.js",
     "migrate": "$npm_package_config_grunt db:migrate:up",
-    "check-migrate": "$npm_package_config_grunt db:migrate:up --dry-run | sed -rn 's/.*Processed migration (.*)/\\x1b[41m[WARNING] Pending Migration: \\1\\x1b[0m (Run `npm run migrate`)/p'",
+    "check-migrate": "$npm_package_config_grunt db:migrate:up --dry-run | perl -ne 's/.*Processed migration (.*)/\\x1b[41m[WARNING] Pending Migration: \\1\\x1b[0m (Run `npm run migrate`)/ && print'",
     "init": "npm run migrate && node test/init/init/init.test.js",
     "demo": "npm run migrate && mocha test/test.upstart.js --development --recursive test/demo",
     "import": "cp -v $npm_package_config_dir/exclude.txt exclude.txt | true && rsync -av --exclude-from=exclude.txt $npm_package_config_dir/* .",
--- a/package.json
+++ b/package.json
@@ -66,10 +66,13 @@
   "devDependencies": {
     "casper-chai": "^0.2.1",
     "casperjs": "1.1.0-beta3",
+    "db-migrate": "0.10.0-beta.11",
+    "db-migrate-pg": "0.1.9",
     "mocha-casperjs": "0.5.3",
     "nodemon": "^1.3.7",
     "pg": "~3.6.0",
-    "phantomjs": "^1.9.16"
+    "phantomjs": "^1.9.16",
+    "sails-db-migrate": "1.3.1"
   },
   "config": {
     "grunt": "./node_modules/sails/node_modules/.bin/grunt",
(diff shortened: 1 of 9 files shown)


cwarden:
GitHub/Travis CI aren't showing the [build status for edac7fa](https://travis-ci.org/cwarden/openopps-platform/builds/116692951), probably because I ran it against my fork of the repo before opening the pull request.

ultrasaurus:
Thanks @cwarden -- good to see it is green!

package.json
```diff
@@ -77,18 +80,19 @@
...
     "test:unit": "mocha test/test.upstart.js --recursive test/unit",
     "test": "npm run test:unit && npm run test:api && npm run test:browser",
     "build": "$npm_package_config_grunt build",
-    "start": "node app.js",
-    "migrate": "./tools/postgres/init.sh",
-    "init": "node test/init/init/init.test.js",
-    "demo": "mocha test/test.upstart.js --development --recursive test/demo",
+    "start": "npm run check-migrate && node app.js",
+    "migrate": "$npm_package_config_grunt db:migrate:up",
+    "check-migrate": "$npm_package_config_grunt db:migrate:up --dry-run | sed -rn 's/.*Processed migration (.*)/\\x1b[41m[WARNING] Pending Migration: \\1\\x1b[0m (Run `npm run migrate`)/p'",
```
ultrasaurus: `sed -r` doesn't work on Mac OSX (which I'm running) -- `sed -E` seems to work for me, but I think it won't work on Linux.  http://stackoverflow.com/questions/6761361/sed-regex-problem-on-mac-works-fine-on-linux 

For now I'll continue to test with -E locally... but we'll need to address this somehow.

ultrasaurus:
thanks @cwarden I will verify that works for me.

I've tested this locally, resetting my database, and testing the various commands and it all seems to function in terms of doing the things to the database that I expect!  I did see some things that I had some questions about:

- can we make it so check-migrate doesn’t output all of the pending migrations?  (At first I feared that it was applying all the migrations, but after close reading I saw it hadn't.  I think it would be more friendly to new devs if it didn't echo all of the pending migrations and it would probably be a bit faster too)
- is there any reason to keep the schema table?  I think that was just used for the old psql migration system.
- could we do the check in npm start with just a db query or something a little faster? It looks like everything gets loaded 2x, which is undesirable.  I don't want to create a bunch of wonky code to maintain, but neither do I want to make startup much, much slower... thoughts?

I wrote a very basic migration, and my experience was inconsistent with the docs -- I realize they aren't our docs, but wondering if it might indicate something else is not quite right...

when I generated the migration it looked like this (which was different from the docs):
```
exports.up = function(db) {
  return null;
};

exports.down = function(db) {
  return null;
};
```

I needed to add the `callback` parameter to make it work:
```
exports.up = function(db, callback) {
  db.createTable('things', {
    id: { type: 'int', primaryKey: true },
    name: 'string'
  }, callback);
};

exports.down = function(db, callback) {
  db.dropTable('things', callback);
};
```

Not a big deal if that's needed, I just like to understand why, and if it is expected maybe add a note in our docs.

Thanks!

cwarden:
On Fri, Mar 18, 2016 at 04:07:59PM -0700, Sarah Allen wrote:
> - can we make it so check-migrate doesn’t output all of the pending
> migrations?  (At first I feared that it was applying all the
> migrations, but after close reading I saw it hadn't.  I think it
> would be more friendly to new devs if it didn't echo all of the
> pending migrations and it would probably be a bit faster too)

What would you like check-migrate to output instead?

Since the application is likely to be broken until the migrations are
run, and new devs are least likely to know why it's broken, I would
err on the side of making sure the developer knows there are
migrations that need to be run.

> - is there any reason to keep the schema table?  I think that was
> just used for the old psql migration system.

I don't think the old schema needs to be kept except that
`tools/postgres/init.sh` references it and the requirements mentioned
that there could be developers who need to run the old migrations.

If you only want to support developers on schema version 12 or those
starting with blank slate, we could delete `init.sh`, `current.sql`,
and all of the old migration shell scripts.

> - could we do the check in npm start with just a db query or
> something a little faster? It looks like everything gets loaded 2x,
> which is undesirable.  I don't want to create a bunch of wonky code
> to maintain, but neither do I want to make startup much, much
> slower... thoughts?

Using the grunt tasks (which load the Sails config, then shell out to
db-migrate) is the documented interface for sails-db-migrate.

Perhaps future enhancements could be to develop a Sails hook that
that calls db-migrate during initialization to check for pending
migrations, or to contribute changes to db-migrate to expose an API
that sails-db-migrate could use rather than shelling out.  I think the
sails initialization to load the configuration is the primary
overhead.

I wouldn't recommend querying the database directly as you would
probably be re-implementing two private APIs, to get the sails
database configuration and to query the db-migrate managed
`migrations` table and compare the results with what's in the
file system.

> I wrote a very basic migration, and my experience was inconsistent
> with the docs -- I realize they aren't our docs, but wondering if it
> might indicate something else is not quite right...
> 
> when I generated the migration it looked like this (which was different from the docs):
> ```
> exports.up = function(db) {
>   return null;
> };
> 
> exports.down = function(db) {
>   return null;
> };
> ```
> 
> I needed to add the `callback` parameter to make it work:
> ```
> exports.up = function(db, callback) {
>   db.createTable('things', {
>     id: { type: 'int', primaryKey: true },
>     name: 'string'
>   }, callback);
> };
> 
> exports.down = function(db, callback) {
>   db.dropTable('things', callback);
> };
> ```
> 
> Not a big deal if that's needed, I just like to understand why, and
> if it is expected maybe add a note in our docs.

I noticed that as well, which is why I made sure to link to the
documentation on how to write a migration.  The template was changed
during the 0.10 beta release cycle:
https://github.com/db-migrate/node-db-migrate/commit/5eb2b6d633d18e4fead8d61875e0075a057f59fa
0.10 has been in beta since October 2015.

ultrasaurus:
sounds reasonable.  I'll just check your latest change locally to make sure it runs fins on mac.

Ideally check-migrate would just output the names of the pending migrations.  Not a show stopper...just feeling like our console output is a bit verbose in places.

ultrasaurus:
everything works on my Mac OSX laptop.  I even tested upgrading from a v11 schema and it all seems to work.  I will merge and I believe that will auto-deploy to: https://openopps-test.18f.gov/   so we'll test it out with a live deployment.

cwarden:
Here's a prototype for a future enhancement using a Sails hook to check for pending migrations:
https://github.com/cwarden/openopps-platform/commit/86061676ccdd603ff4af96ac77f200ee147344ed

It requires an upgrade to Sails and a few small changes to sails-db-migrate, so it's definitely not ready to be merged, but as a proof-of-concept, it looks like a good approach.

PR #1300, opened 2016-04-26
Logged in dasboard and user creation script

csarsam:
Summary
---

1. Re-add the `ActivityController` so that the logged in dashboard retrieves the data it needs successfully.
2. Re-add a div to the task view template so that it can be hidden successfully during task editing.
3. Add a script, `demo.js`, that adds a csv file of users to the database when `npm run demo` is run.

How to Test
---

1. There should be no errors on the logged in dashboard
2. There should be no non-editable task view at the bottom of the edit task view
3. After `npm run demo` is run, there should be 12 users in the database.

---

FYA @ultrasaurus


ultrasaurus:
FYI: tests pass locally -- travis still configured with old version of node

ultrasaurus:
rebased and merged locally, tested:
- no errors on the logged in dashboard
-  no non-editable task view at the bottom of the edit task view
- After npm run demo is run, there should be 12 users in the database. --> the users.csv was ignored, fixed gitignore & error reporting in another commit

Thanks @csarsam !

PR #1312, opened 2016-05-08
Volunteer test - small refactor

ultrasaurus:
small refactor of https://github.com/18F/openopps-platform/pull/1311

@csarsam what do you think?

### DIFF
--- a/test/integration/models/Volunteer.test.js
+++ b/test/integration/models/Volunteer.test.js
@@ -1,0 +1,127 @@
+var assert = require('chai').assert;
+var users = require('../../fixtures/user');
+var sinon = require('sinon');
+var Promise = require('bluebird');
+
+// Reusable utilities
+
+function createUsers(numUsers) {
+  var promises = [];
+
+  for (var i = 0; i < numUsers; i++) {
+    promises.push(User.create(
+      {
+        'name': i.toString(),
+        'username': i.toString() + '@gmail.com',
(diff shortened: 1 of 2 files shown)


csarsam:
@ultrasaurus Your improvements look good to me - the creatVolunteers utility definitely needed the comment explaining why the creation needs to be serial. (I'd like to know why `findOrCreate` doesn't do what it says on the box when the requests are made at the same time, but it does seem safe to assume that the test timing conditions wouldn't be possible under real usage). I didn't know you could do creation of models using an array of attributes either, that's definitely cleaner.

test/integration/models/Volunteer.test.js
```diff
@@ -0,0 +1,134 @@
...
+      }
+    );
+  }
+  return User.create(userAttrs);
+}
+
+function createVolunteers(users, taskId) {
+  // creating 4 volunteers at once, creates 4 badges, not sure why...
+  // but the app never actually does this
+  // var volAttrs = [];
```
csarsam: This commented out code should get cleaned up.

csarsam:
@ultrasaurus Should this get re-opened against `dev` now that `dev-new` has gotten merged?

ultrasaurus:
yep!  I'll see if I can get this done today for 0.10.1

ultrasaurus:
closing in favor of: https://github.com/18F/openopps-platform/pull/1327

PR #1321, opened 2016-05-14
Correct task creator id field references in TaskController.export

csarsam:
**Problem**

Attempting to perform a task export resulted in a `500`.

**Solution**

Update outdated property references in `TaskController.export`.

**How to +10/QA**

1. `$ git checkout task-export && npm start`
2. Navigate to [http://localhost:1337/admin/tasks](http://localhost:1337/admin/tasks) and click on the 'download' icon.
3. A .csv file populated with the correct data should download.

---

FYA @ultrasaurus @DavidEBest

### DIFF
--- a/api/controllers/TaskController.js
+++ b/api/controllers/TaskController.js
@@ -129,7 +129,7 @@ module.exports = {
       }
       sails.log.debug('task export: found %s tasks', tasks.length)
 
-      User.find({id: _.pluck(tasks, 'userId')}).populate('tags').exec(function (err, creators) {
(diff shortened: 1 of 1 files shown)


ultrasaurus:
thanks @csarsam !  I opened a bug for adding a test:  https://github.com/18F/openopps-platform/issues/1323  but I'm glad you went ahead and submitted the fix since I had included this with the 0.10.0 release.  I'll tag with 0.10.1 and prolly do a bug fix release later today with this and the theme support

PR #1322, opened 2016-05-15
changes required to make the theme work

ultrasaurus:
changes both here and in the theme itself to make the skinning / config changes work for the federal instance 
also added analytics from 0.9.5

testing:
in the directory above openopps-platform:
```
git clone git@github.com:18F/open-opportunities-theme.git
cd openopps-platform
npm run import --openopps-platform:dir=../open-opportunities-theme
```


ultrasaurus:
tested manually on staging server -- this seems to work

PR #1331, opened 2016-05-15
use .sailsrc to turn off sockets

### LINKED ISSUE #1330: as an admin, I should not see so many errors on the server (turn off sails sockets)
would be fun to do something with them
but since we don't use 'em yet, we should configure to be off

note: since the servers aren't configured to support them, they throw all sorts of errors really frequently!

ultrasaurus:
in the layouts, conditionally turn off sockets
- when testing
- when configured to be off

following guidance here: https://github.com/balderdashy/sails/issues/3580#issuecomment-185089243

resolves #1330

### DIFF
--- a/.sailsrc
+++ b/.sailsrc
@@ -1,0 +1,6 @@
+{
+  "hooks": {
+    "sockets": false,
+    "pubsub": false
+  }
+}
--- a/views/layout.ejs
+++ b/views/layout.ejs
@@ -87,10 +87,7 @@
 
     -->
 
-    <!--SCRIPTS-->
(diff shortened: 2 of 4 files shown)


ultrasaurus:
would be nice to have another pair of eyes on this... @csarsam @DavidEBest --- if I'm reading the notes on https://github.com/balderdashy/sails/issues/3580#issuecomment-185089243 correctly, then the client-side JavaScript for SailsJS is *only* used for sockets and not for anything else we need.

ultrasaurus:
I've tested this a bit and it seems to work, so will merge to include in 0.10.1

PR #1334, opened 2016-05-16
Addressing a few comments.

DavidEBest:
In my short-sightedness, I started this work on a branch instead of a fork.


ultrasaurus:
@DavidEBest thanks so much for addressing the comments -- looks great!

PR #1348, opened 2016-05-19
check for null owners

ultrasaurus:
tasks should never have a null owner,
but if they do, just log to the console
and still display everything else


ultrasaurus:
this template is yearning to be refactored, but not today :(

Issue #1299, opened 2016-04-22
As a GSA employee, I would like to see GSA Opportunities on a page that feels like a "place"


ultrasaurus:
As a General Services Administration (GSA) employee, I would like a page that can be shared that will take people to a “GSA Only” page that lists just all GSA opportunities (part time and full time) so I don’t have to search all opportunities in order to find just the GSA ones, and I feel like this is a GSA place. As the program administrator for the Office of Human Resource Management (OHRM), this will enable us to send out program announcements and easily point employees to the GSA-specific opportunities.

This page would not require login (so technically people at other agencies would be able to see it if a link were shared, but they wouldn't typically be presented with this page).

This page would have a specific URL that OHRM could send out to GSA employees that would take participants to GSA only rotation program only (maybe including GSA part-time projects too, maybe promote federal-wide in some way)

ultrasaurus:
Here are some mocks from design ideation sessions with @vz3 and @brendansudol 

1 - "Streamlined Header"
![image](https://cloud.githubusercontent.com/assets/41906/14747058/50f70f88-0867-11e6-8be0-09912b40b31b.png)

2 - "Branded Sidebar"
![image](https://cloud.githubusercontent.com/assets/41906/14747118/8c0379d6-0867-11e6-8c53-41198e176905.png)

3 - "Create Opportunity Highlight"
![image](https://cloud.githubusercontent.com/assets/41906/14747142/a46b7122-0867-11e6-97fc-bcc14a46aaa7.png)

ultrasaurus:
Note: upon review by GSA OHRM team, they liked "2 - Branded Sidebar" option best, feeling it was cleaner and they liked the way the gray background helped the opportunities to stand out

mtorres253:
This seems like a duplicate of #1271

Issue #1225, opened 2016-01-27
As a program administrator, I should be able to see an admin tasks page for my agency


ultrasaurus:
With a URL like:  /admin/tasks/:agencyId 
I should see a page like this:
![image](https://cloud.githubusercontent.com/assets/41906/14511417/3d62b7a6-018c-11e6-9977-8cf2de93c433.png)

showing only tasks created by users from my agency

depends on: https://github.com/18F/openopps-platform/issues/1194

PR #1349, opened 2016-05-19
Agency Admin user experience

ultrasaurus:
#1299 As a program administrator, I should be able to see all of the users from my agency
#1225 As a program administrator, I should be able to see an admin tasks page for my agency

## Interactive Testing Notes

```
node tools/postgres/addGSAdomain.js
```

in browser, go to : http://localhost:1337/ and make an account with agency GSA
then on the command line

```
psql midas
update midas_user set "isAgencyAdmin"='t' where username='newuser@whatever.com';
```

## Implementation Notes
added agency meta-data (abbr & domain)
/admin/agencies/:abbr
/admin/users/:abbr
/admin/tasks/:abbr

utility script to add meta-data to GSA
  {abbr: "GSA", domain: "gsa.gov"}

agency admin page
use agency abbreviation for slug
- /admin/agencies/:abbr
- display agency meta-data
admin/tasks/:abbr
admin/users/:abbr
- consistent naming for ‘admin/users’ 
  (was admin/user in some places)

create a client-side User object 
- easily access agency data

Server-side
- User.findByDomain
- AdminController users id:gas => filter by gsa.gov

## Automated testing notes
- move createUsers into modelUtils 
- added createTasks to modelUtils
- 2 new tests marked pending, since they require setup 
  - User.findByDomain 
  - Task.findByOwnerDomain

  they use SQL query so require postgres DB  to run them create a new DB midastest and set up with the schema, but no data (added note to CONTRIBUTING.md for how)

### DIFF
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -456,7 +456,6 @@ module.exports = {
   * eg: /api/admin/tasks
   */
   tasks: function ( req, res ) {
-    sails.log.verbose('AdminController tasks', req.allParams())
     var page = parseInt( req.param( 'page', 1 ) ),
       limit = req.param( 'limit', 1000 ),
       sort = req.param( 'sort', 'createdAt desc' ),
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -477,7 +477,6 @@ module.exports = {
     }
 
     findTasks.then(function(output) {
-      console.log('output ---->', output)
       res.json( output );
     })
     .catch(function(err) {
--- a/api/controllers/AdminController.js
+++ b/api/controllers/AdminController.js
@@ -21,44 +21,62 @@ module.exports = {
    * @param q the filter to apply to the search (filters by name)
    */
   users: function (req, res) {
+    sails.log.verbose('AdminController users', req.allParams())
     var page = parseInt(req.param('page', 1));
     var limit = req.param('limit', 25);
     var query = req.param('q');
-    var where = {};
+    var id = req.param('id');
 
-    if (query) {
-      // search by both name and username
-      where = {
-        or: [
-          { like: { name: '%' + query + '%' }},
-          { like: { username: '%' + query + '%' }}
-        ]
-      };
-    }
+    if (id) {  // this will be the agency id, e.g. 'gsa'
+      var domain = id + '.gov';   // just prototyping, should refactor to use data model
+      User.findByDomain(domain)
+      .then(function(users) {
+        return res.send({
+          page: 1,
+          limit: users.length,
+          count: users.length,
+          users: users,
+          q: query
+        });
+      })
+      .catch(res.negotiate);
+    } else {
+      var where = {};
 
-    // find users that meet this criteria
-    User.find()
-      .where(where)
-      .sort({ createdAt: 'desc' })
-      .paginate({ page: page, limit: limit})
-      .exec(function (err, users) {
-        if (err) { return res.send(400, { message: 'Error looking up users', err: err}); }
-        // count the total number of users
-        User.count(function (err, count) {
-          if (err) { return res.send(400, { message: 'Error counting users', err: err}); }
-          async.each(users, addUserMetrics, function(err) {
-            if (err) { res.send(400, { message: 'Error retrieving metrics for users.', err: err}); }
-            // return a paginated object
-            return res.send({
-              page: page,
-              limit: Math.min(users.length, limit),
-              count: count,
-              users: users,
-              q: query
+      if (query) {
+        // search by both name and username
+        where = {
+          or: [
+            { like: { name: '%' + query + '%' }},
+            { like: { username: '%' + query + '%' }}
+          ]
+        };
+      }
+
+      // find users that meet this criteria
+      User.find()
+        .where(where)
+        .sort({ createdAt: 'desc' })
+        .paginate({ page: page, limit: limit})
+        .exec(function (err, users) {
+          if (err) { return res.send(400, { message: 'Error looking up users', err: err}); }
+          // count the total number of users
+          User.count(function (err, count) {
+            if (err) { return res.send(400, { message: 'Error counting users', err: err}); }
+            async.each(users, addUserMetrics, function(err) {
+              if (err) { res.send(400, { message: 'Error retrieving metrics for users.', err: err}); }
+              // return a paginated object
+              return res.send({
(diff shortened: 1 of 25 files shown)


csarsam:
@ultrasaurus When I try to load the http://localhost:1337/admin/agencies page, I see an error - 'Uncaught TypeError: Cannot read property 'abbr' of null'. Looking at the `agencyTag` variable, `agencyTag.data` is `null`. Is there any setup I need to be doing for this?

ultrasaurus:
@csarsam um... I need to work on my PR notes, been working solo too long!   I totally forgot to write down that this feature requires some data, added with the script

`node ./tools/postgres/addGSAdomain.js`

(which then runs the server and you need to stop with ctrl-c)

Then if you add GSA as your agency and you are an admin or agency admin, you will see the new UI.

Though I totally forgot to test the normal case... maybe the agency dashboard tab should just be hidden in that case.

api/controllers/AdminController.js
```diff
@@ -438,7 +456,7 @@ module.exports = {
   * eg: /api/admin/tasks
   */
   tasks: function ( req, res ) {
-
+    sails.log.verbose('AdminController tasks', req.allParams())
```
DavidEBest: Can this be removed?

api/controllers/AdminController.js
```diff
@@ -21,44 +21,62 @@ module.exports = {
...
-        or: [
-          { like: { name: '%' + query + '%' }},
-          { like: { username: '%' + query + '%' }}
-        ]
-      };
-    }
+    var id = req.param('id');
+
+    if (id) {  // this will be the agency id, e.g. 'gsa'
+      var domain = id + '.gov';   // just prototyping, should refactor to use data model
```
DavidEBest: What doest this comment mean? What should be done to use a data model here?

api/controllers/AdminController.js
```diff
@@ -449,9 +467,17 @@ module.exports = {
       },
       openTasks,
       steps = [];
+    var id = req.param('id'); // this will be the agency id, e.g. 'gsa'
+    var domain;
+    if (id) {
+      domain = id + '.gov';   // just prototyping, should refactor to use data model
```
DavidEBest: Same as the above. Can this reference the same data model methods?

api/controllers/AdminController.js
```diff
@@ -449,9 +467,17 @@ module.exports = {
...
+      findTasks = Task.byCategoryForDomain(domain)
+    } else {
+      findTasks = Task.byCategory(page, limit, sort, domain)
+    }
 
-    Task.byCategory(page, limit, sort)
-    .then(function(output) {
+    findTasks.then(function(output) {
+      console.log('output ---->', output)
```
DavidEBest: Debug output

api/models/Task.js
```diff
@@ -188,37 +189,64 @@ module.exports = {
...
           .sort(sort)
           .paginate({ page: page, limit: limit })
       .then(function(allTasks) {
         tasks = allTasks;
-        volQuery = _.map(tasks, function(t) { return {taskId: t.id} });
-        return Volunteer.findUsersByTask(volQuery)
+        return Task.byCategoryWithVolunteers(tasks);
       })
+      .then(function(output) {
+        console.log('here')
```
DavidEBest: More debug logging

api/models/Task.js
```diff
@@ -269,6 +297,19 @@ module.exports = {
     return promise;
   },
 
+  // find where owner has username with matching domain
+  // returns a promise
+  findByOwnerDomain: function(domain) {
+    var queryStr = "SELECT task.* from task where \"userId\" " +
```
DavidEBest: Is it possible to do this with the orm? If not we should at least parameterize this query.

api/models/User.js
```diff
@@ -174,6 +174,17 @@ module.exports = {
 
   },
 
+  // returns a promise
+  findByDomain: function(domain) {
+    var queryStr = "SELECT * FROM midas_user WHERE username SIMILAR TO '%(@|.)" +
```
DavidEBest: As above, should be parameterized or handled by the orm.

assets/js/backbone/apps/admin/views/admin_agencies_view.js
```diff
@@ -8,35 +8,46 @@ var AdminAgenciesTemplate = fs.readFileSync(`${__dirname}/../templates/admin_age
...
   },
 
   initialize: function (options) {
     this.options = options;
+    this.adminMainView = options.adminMainView;
     this.data = {
-      agency: _(window.cache.currentUser.tags).findWhere({ type: 'agency' })
+      agency: window.cache.currentUser.agency
     };
+    console.log('AdminAgenciesView data', this.data.agency);
```
DavidEBest: Debug output?

ultrasaurus:
on the comment about prototyping an agency data model..

My thinking is to see if the UI actually works for GSA, then create a Sails model to represent agencies.  Many agencies have multiple domains (usda has at least a dozen for just the folks who use Open Opps -- OMB is generally considered to be an "agency" though technically it is an "office" that is part of Executive Office of the President (omb.eop.gov), same with OPM (Office of Personnel Management) but they have their own domain (opm.gov).  My current thinking is we probably want to implement an "org" model that can have a number of domains, which could broaden the use case to include things like Code for Dayton being an org as part of a larger civic Open Opps instance (someday, maybe).  However, in the short-term, I was thinking it would be helpful to do a full implementation with our first program users, make sure they can use it successfully, then refactor.

@DavidEBest thoughts?

api/controllers/AdminController.js
ultrasaurus: see comment below... it's a pretty significant change to create a data model for an agency (rather than using the current tag system)... long overdue, but seems like we should validate UX first

api/controllers/AdminController.js
ultrasaurus: see comment below... it's a pretty significant change to create a data model for an agency (rather than using the current tag system)... long overdue, but seems like we should validate UX first

api/models/Task.js
ultrasaurus: The ORM doesn't handle SIMLAR query and I couldn't figure out any Sails/Waterline way to do query parameterization, though it seems like it *should* be available

ultrasaurus:
closing in favor of: https://github.com/openopps/openopps-platform/pull/1359

Issue #1276, opened 2016-03-29
When browsing the website, when I view a full-time detail, I should not see a time commitment


ultrasaurus:
This happened when editing the full-time detail, switching to "on-going" then saving, then switching back....

It should never look like this:
![image](https://cloud.githubusercontent.com/assets/41906/14115091/78802990-f58e-11e5-80d8-04eaf1395e7e.png)

It should show just "Full time detail"  not "— up to 2 hours per week"

ultrasaurus:
"Time Committment" shouldn't have two t's

PR #1355, opened 2016-05-26
Don't render time commitments for full-time details

csarsam:
*Problem*
See issue #1276 

*Solution*
Add a conditional in the rendering template that checks for whether the required time commitment is full time, and if so doesn't  render the time commitment estimate.

@ultrasaurus Should we not show the ' 'How much time is needed?' option for  when editing full-time detail tasks? I could stick that in this PR too.


ultrasaurus:
@csarsam looks good!  Yep, there's another issue on that:  https://github.com/openopps/openopps-platform/issues/1275

PR #1362, opened 2016-06-05
temporarily remove browser tests

ultrasaurus:
they haven’t actually worked since 0.10.0 with the big upgrade
the client files cause Phantom “Parse Error” with no stack trace
Phantom also takes a long time to install 
& confusing to have code that doesn’t actually work


ultrasaurus:
FYI: I pushed the removed tests in a separate branch in case someone wants to keep working on them:  https://github.com/openopps/openopps-platform/tree/test-browser  (see https://github.com/openopps/openopps-platform/issues/1363)

DavidEBest:
It's a bummer that they needed to be removed, but 👍 . Good call on creating issue #1363 so they don't stay removed. :)

PR #1365, opened 2016-06-07
Re-adding a Dockerfile to the repo.

DavidEBest:
This seems to have gotten removed at some point. Updated the node version to a more recent release, but otherwise unchanged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM node:4.4.5-onbuild
+
+# Expose environment variables
+ENV BRANCH= \
+  NEW_RELIC_APP_NAME= \
+  NEW_RELIC_LICENSE_KEY= \
+  LINKEDIN_CLIENT_ID= \
+  LINKEDIN_CLIENT_SECRET= \
+  LINKEDIN_CALLBACK_URL= \
+  MYUSA_CLIENT_ID= \
+  MYUSA_CLIENT_SECRET= \
+  MYUSA_CALLBACK_URL= \
+  MYUSA_CLIENT_ID= \
+  MYUSA_CLIENT_SECRET= \
+  AWS_ACCESS_KEY_ID= \
+  AWS_SECRET_ACCESS_KEY= \
+  EMAIL_HOST= \
+  EMAIL_USER= \
+  EMAIL_PASS= \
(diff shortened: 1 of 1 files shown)


ultrasaurus:
I'm pretty new to docker, and didn't know how to test it, so left it out.  Thanks so much for digging it up!  I couldn't find reference docs anywhere... would you mind adding a simple doc/dev/docker.md or wherever you think it makes sense to reference how to use this?  (I assume that this is for us to create a docker instance for DockerHub or for someone to use for AWS deployment?)

Thanks!
Sarah

DavidEBest:
Will do. Turns out, I have a bit of work ahead of myself for this still. The last time 18F had a docker image of this app was version 0.7.3, which did not have a hard dependency on postgresql.

ultrasaurus:
Maybe we should prioritize refactoring the agency sub-organization feature so it doesn't need postgres...

DavidEBest:
I've dug into this a bit, and I think we might be stuck with postgres. The migrations require it, or something like it.

[db-migrate](https://github.com/db-migrate/node-db-migrate) has support for postgres, mysql, and sqlite.
[sails-db-migrate](https://github.com/building5/sails-db-migrate) [explicitly limits](https://github.com/building5/sails-db-migrate/blob/4df349bf3779d4d1e072914663bd8de537800007/lib/sailsDbMigrate.js) support to just postgres & mysql.

There _is_ a [sqlite3 adapter](https://github.com/waterlinejs/sqlite3-adapter) for waterline available, but the sails-db-migrate adapter doesn't support it right now.

The way I see it, there are two options. I can:
* Add sqlite3 support to sails-db-migrate and include sqlite3 in the docker image. It's not as easy as `sails-disk` or `sails-memory`, but it'd be lighter than postgresql.
* Add postgres to the docker image, and update our documentation.

Thoughts, @openopps/coreteam & others?

ultrasaurus:
thanks for researching this @DavidEBest 

Given what you have discoverd, I'd be inclined to require postgres -- even though the setup is a little more time-consuming, it allows us to focus on just a single environment, which will help with our focus and likely improve the quality of what we can deliver.

rogeruiz:
This looks good to me. I can't test the `Dockerfile`, but having it in the repo is a good start on iterating how we use it.

Issue #1293, opened 2016-04-13
As a federal employee, when I browse opportunities, I would like to see which are "virtual"


ultrasaurus:
Right now, we only display a location when a specific location is required.  Virtual opportunities should show as “virtual” because it might be confusing if only the specific location opportunities show a location.

In the image below, the opportunity on the right does not have a location requirement:

![image](https://cloud.githubusercontent.com/assets/41906/14510538/e652fb6a-0186-11e6-869a-913e53c59d8a.png)

Proposal that it appear like this:
![image](https://cloud.githubusercontent.com/assets/41906/14510661/55fc511e-0187-11e6-9429-8f8f80f7ca9d.png)

LisaLNelson:
Completed

PR #1366, opened 2016-06-07
Add 'Virtual' to tasks with no location tag associated

csarsam:
*Problem*

See issue #1293 

*Solution*

Render an entry in the list of info about the task in the browse view clarifying that the task is 'Virtual', ie completable from anywhere


ultrasaurus:
looks good to me :+1:

PR #1367, opened 2016-06-07
show state of checkbox for restricted tasks

ultrasaurus:
fix for https://github.com/openopps/openopps-platform/issues/1361

When a task is restricted to a specific agency
the model has the attribute ‘restrict’ 
set to the agency abbr

To test:

1. run the scripts to set GSA agency data:

  ```
  node ./tools/postgres/addGSAdomain.js
  node ./tools/postgres/addGSAFullTimeDetail.js
  ```
2. set your agency to be GSA in your profile

3. Create a new task, under “What type of effort is needed?” you should see a checkbox with label:
  “Limit Participation to GSA Employees”

the state of the checkbox should be saved and display
with the correct state when the task is edited

### DIFF
--- a/assets/js/backbone/apps/tasks/task-form-view-helper.js
+++ b/assets/js/backbone/apps/tasks/task-form-view-helper.js
@@ -49,7 +49,7 @@ module.exports = {
       timeRequiredTag;
 
     // restrict is a String with agency abbr or empty string for not restricted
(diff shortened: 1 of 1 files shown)


csarsam:
+1

* Tested locally
* Tasks created with the 'Limit Participation' option render in the edit view with the checkbox checked

PR #1368, opened 2016-06-08
fix for when agency doesn’t have data defined

ultrasaurus:
This was throwing errors when the database
wasn’t set up special for GSA (and also would
likely have been an issue for other agencies)


ultrasaurus:
We tested on @csarsam's machine with a user that had a non-GSA agency and it worked fine, but needed to fix the case for a user with no agency (which federal open opps has from older version of the software).

PR #1369, opened 2016-06-08
we shouldn’t depend on task-time-required id

ultrasaurus:
It looks like the code depended on the id of the tag
which was in the database, and that I changed the order
of creation when I re-wrote the scripts that create
the seed data.  Other places in the code it 
used the ‘value’ so I made that change here as well.


csarsam:
+1

* Tested locally
* Works as expected with ongoing and one time tasks

PR #1382, opened 2016-06-15
check for presence of task-length tag

### LINKED ISSUE #1380: Some tasks fail to render
I have 2 assigned tasks in Open Opportunities and I tried to open one and I just get
screen shot 2016-06-14 at 10 24 18 pm
 a blank screen.

ultrasaurus:
should render when task-length tag is missing, 
which is true for some tasks on 18F production/staging

fixes #1380


rogeruiz:
👍

PR #1397, opened 2016-07-03
Site should render on IE and Firefox

### LINKED ISSUE #1392: As a user, I need to be able to open the platform in browsers other than chrome
Several users have expressed problems with opening the platform in various browsers.  Here is the message from Terry Lynn Sales CBP/DHS:

Darn it!  Tried Firefox, Edge, IE and Chrome.  Finally got it to come up in Chrome.  This might limit your respondents if Chrome is the only browser they open in.  I used to be able to open in IE.  Recent change?  

Another message from Johnothan Rears Data Manager/USDA (long time user) cannot open the platform in Firefox or IE and he used to be able to. He also said he has issues with Chrome as well. He can read the posts but you cannot  respond.

Safari does not seem to be working quite right either.  The site will open and tasks appear but not all of them open. For example: [Seeking Content Writer's for DigitalGov blog](url) opens to a blank page.

screen shot 2016-06-17 at 1 16 26 pm

ultrasaurus:
removed jsdom, doesn’t appear to be used
Caused site not to render in IE & Firefox

fixes #1392


package.json
```diff
@@ -116,7 +115,7 @@
     "jshint": "find assets/js/backbone -name \\*js -exec ./node_modules/jshint/bin/jshint {} \\;",
     "cover": "istanbul cover ./node_modules/mocha/bin/_mocha -- test/bootstrap.test.js test/integration/**/*.test.js",
     "debug": "node debug app.js",
-    "browserify": "browserify assets/js/backbone/app.js -o assets/js/bundle.js -t [ babelify --presets [ es2015 stage-0 ] ] -t [ brfs ]",
+    "browserify": "browserify -d assets/js/backbone/app.js -o assets/js/bundle.js -t [ babelify --presets [ es2015 stage-0 ] ] -t [ brfs ]",
```
rogeruiz: Do we want to keep the `-d` debug flag here?

PR #1398, opened 2016-07-03
not-loggedin user should see /

### LINKED ISSUE #1393: as a non-logged in user, when I go to the home page, it should not prompt me to log in
when I go to https://openopps.digitalgov.gov/
I should see the home page (currently it re-directs to the dashboard and we're seeing 403 errors in NewRelic from this)

ultrasaurus:
when not logged in, user should not be re-directed to /dashboard
this was broken when I added User object
forgot to test for null user

fixes #1393

### DIFF
--- a/assets/js/backbone/app-run.js
+++ b/assets/js/backbone/app-run.js
@@ -11,7 +11,7 @@ var Application = {
     var self = this;
     // Cache user
     // Check if a user is already defined
-    if (!_.isUndefined(backendUser)) {
+    if (!_.isUndefined(backendUser) && backendUser != null) {
       window.cache.currentUser = new User(backendUser);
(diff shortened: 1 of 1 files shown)


assets/js/backbone/app-run.js
```diff
@@ -11,7 +11,7 @@ var Application = {
     var self = this;
     // Cache user
     // Check if a user is already defined
-    if (!_.isUndefined(backendUser)) {
+    if (!_.isUndefined(backendUser) && backendUser != null) {
```
rogeruiz: The second check should be enough for this to verify that the `backendUser` is not `undefined` or not `null`. But we can leave them both in for now.

PR #1417, opened 2016-08-02
Changes for easier deployment to Heroku

phildini:
Use case: Attempting to deploy the OpenOpps Platform to the Heroku PaaS.

Changes: 
- Assuring that the two places that pull database config from the cfConf fall back to the `DATABASE_URL` in the environment, which is Heroku (and others) preferred method of setting the database.
- Make theming easier to use with Heroku by removing the `_theme` directory on each install.

Testing: Tested locally and on Heroku.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "dir": "."
   },
   "scripts": {
-    "preinstall": "if [ \"$THEME\" ]; then git clone $THEME _theme; npm run import --openopps-platform:dir=_theme; fi",
+    "preinstall": "if [ \"$THEME\" ]; then rm -rf _theme; git clone $THEME _theme; npm run import --openopps-platform:dir=_theme; fi",
     "import": "cp -v $npm_package_config_dir/exclude.txt exclude.txt | true && rsync -av --exclude-from=exclude.txt $npm_package_config_dir/* .",
     "install": "npm run migrate && npm run uswds && node --version && pwd && ls",
     "init": "node ./tools/init.js",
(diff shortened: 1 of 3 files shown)


rogeruiz:
Thanks for the contribution @phildini. I will get to this in the morning, but it seems like these are good things to include in the platform. Thanks again. 👍

phildini:
Any update here? Thanks again for your time. :)

package.json
```diff
@@ -102,7 +102,7 @@
     "dir": "."
   },
   "scripts": {
-    "preinstall": "if [ \"$THEME\" ]; then git clone $THEME _theme; npm run import --openopps-platform:dir=_theme; fi",
+    "preinstall": "if [ \"$THEME\" ]; then rm -rf _theme; git clone $THEME _theme; npm run import --openopps-platform:dir=_theme; fi",
```
rogeruiz: Is the problem here that the `_theme/` directory is being cached on the server which is running `npm install`? I don't see anything particularly bad with removing the `_theme/` directory. I'm curious to hear the reasoning behind removing the directory.

rogeruiz:
Thanks @phildini I added a single line comment, but this looks good once I hear back. Thanks!

package.json
phildini: This is more for local development then heroku development, but is of great assistance to the people who are working with on getting this done for heroku. 

The logic here is that right now you need to manually remove the `_theme` directory if you want to get new changes from updates to the repo. Otherwise, if you've already run `preinstall` before, you get: 

`fatal: destination path '_theme' already exists and is not an empty directory`

PR #1439, opened 2016-08-24
[ work in progress ] Initial commit for bringing the map back;

rogeruiz:
This patch updates the codebase to render a map, but it isn't working
properly. @EricSchles will be :eyes: looking into this :eyes:.


rogeruiz:
Closing this for now. We won't be rendering a map on the `peoples` tab for now.

All open source versions are too complex and/or require having a 3rd-party SaaS access account to create a map.

Issue #1404, opened 2016-07-12
As an admin I would like to be able to change the name of the owner of a task. 


LisaLNelson:
This is something that worked until recently. A task creator was unable to add an opportunity so I said I would do it for them.  I was able to add the opportunity but when I tried to change the name of the requestor from me to the person i who wanted, the tasks gets created but it would not save.  It would remain in drafts in my name.  I tried it several times both in IE and in Chrome.

## Acceptance Criteria 
- [x] Admins can edit a task's creator 
- [x] Admins can save the task after editing a task creator's name

rogeruiz:
Reports the following in the JavaScript console

<details>
<summary>Error happening in bundle.js around line within `#newTask#template` </summary>
```txt
TypeError: undefined is not an object (evaluating 'agency.abbr')
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/601.6.17 (KHTML, like Gecko) Version/9.1.1 Safari/601.6.17
```
</details>

<img width="418" alt="screen shot 2016-07-13 at 2 21 05 pm" src="https://cloud.githubusercontent.com/assets/706004/16814786/57f72d20-4905-11e6-9135-47afed2999e1.png">

rogeruiz:
The original issue was that the page was not rendering, but on further inspection the database is not being updated here.

brodydittemore:
Is this issue related to #1500 and #1529?

brodydittemore:
Can verify that this works in Chrome as of 8/31/17.
Have not verified in IE.

brodydittemore:
Verified that this works in IE11 as of 8/31/17.

PR #1440, opened 2016-08-25
Pass owner user model to modelData

rogeruiz:
This patch adds the owner directly to the modelData being passed to the
server.


rogeruiz:
This fixes #1404.

PR #1446, opened 2016-08-30
Fix the developer workflow

rogeruiz:
This patch contains fixes to the developer workflow by adding BrowserSync and
using Grunt to provide the Browserify support. The project now properly supports
the watch workflow, but still takes a long time to reload the page.


rogeruiz:
SailsJS uses Grunt to process workflows, so this patch series does that instead of relying on command-line tools.

PR #1499, opened 2017-02-09
Open Opportunities deployable on GovCloud

rogeruiz:
🚧 https://openopps.app.cloud.gov/

Migrating from cloud.gov East / West to :cloud:.gov GovCloud.
cc @LisaLNelson

AC:

- [x] Migrate data from EW to FR ☁️.gov
- [x] Migrate production app from EW to FR ☁️.gov
- [x] Migrate manual deployment from EW to FR ☁️.gov
- [ ] ~~Migrate automated deployment from EW to FR ☁️.gov~~
- [ ] ~~Update the SSP with updated info pointing to the new environment~~
- [x] Update the codebase to support FR :cloud:.gov

PR #1508, opened 2017-06-07
[WIP] Add preliminary docker-compose support.

toolness:
Note that this PR is against #1499, since that seems to be the latest branch right now.

## Instructions

* Run `docker-compose up db` and wait until you see the log message `PostgreSQL init process complete`.  This means that the database has been created.  Then press ctrl-c.
* Run `docker-compose run app npm install --unsafe-perm`.  The `--unsafe-perm` is needed because you're running some startup scripts as root in the container, which `npm` doesn't like. We can fix that in a later PR though.
* Run `docker-compose run app npm run init`.
* Run `docker-compose up` to start everything and visit http://localhost:3000/.

## Notes

* The docs claim postgres 9.2 is required, but some scripts use features introduced in 9.3.
* The docs are very confusing about what version of node is required, so I just used what was specified in `package.json`.

## To do

- [ ] Add setup instructions
- [x] Make sure tests run in the container

### DIFF
--- a/api/models/Notification.js
+++ b/api/models/Notification.js
@@ -10,6 +10,15 @@
      protocol = sails.config.emailProtocol,
      transportConfig = sails.config[protocol.toLowerCase()];
 
+if (protocol == '') {
+  // We're running tests, just use nodemailer's stream transport.
+  transportConfig = {
+    streamTransport: true,
+    newline: 'unix',
+    buffer: true
+  };
+}
+
 module.exports = {
 
   attributes: {
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,12 +16,11 @@ services:
     environment:
       - PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/app/node_modules/.bin
       - DATABASE_URL=postgres://midas@db/midas
-      - DATASTORE=local
+      - DATASTORE=postgresql
       - LOG_LEVEL=silly
       - PORT=3000
       - SAILS_SECRET=just_for_testing_yo
       - 'VCAP_APPLICATION={ "uris": [ "openopps-test.18f.gov" ] }'
-      - RUNNING_IN_DOCKER_FOR_DEVELOPMENT=yup
   db:
     image: postgres:9.3
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,34 @@
+version: '2'
+services:
+  app:
+    build:
+      dockerfile: Dockerfile.simple
+      context: .
+    volumes:
+      - .:/app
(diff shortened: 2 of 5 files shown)


api/models/Notification.js
```diff
@@ -10,6 +10,15 @@
...
      transportConfig = sails.config[protocol.toLowerCase()];
 
+if (protocol == '') {
+  // We're running tests, just use nodemailer's stream transport.
+  transportConfig = {
+    streamTransport: true,
+    newline: 'unix',
+    buffer: true
+  };
+}
```
toolness: previously when the protocol was an empty string (during testing), it meant `transportConfig` was undefined, which threw an exception when it was passed to nodemailer on my system. So this fixes that.

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    command: npm run watch
+    ports:
+      - 3000:3000
+    links:
+      - db
+    environment:
+      - PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/app/node_modules/.bin
+      - DATABASE_URL=postgres://midas@db/midas
+      - DATASTORE=postgresql
+      - LOG_LEVEL=silly
```
toolness: Yes, "silly" is actually a [legit log level](http://sailsjs.com/documentation/concepts/logging) in sails.

toolness:
I'm not entirely sure where to add instructions for this.  The `INSTALL.md` already has a whole ton of different ways of installing the app, most of which seem to be out of date in some major way...

docker-compose.yml
EricSchles: lol

EricSchles:
```
xcoh2j-19wfvh8:openopps-platform ericschles$ docker-compose up db
ERROR: 
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
        
```
@toolness I tried following your steps and failed on the first one.

EricSchles:
Got error: `ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`  You should probably specify this as the first step in the instructions?  Rather than assuming docker is already running?  @toolness

EricSchles:
Okay, so I'm not sure if this is specific to my machine but after much investigation, I found docker - it doesn't map to localhost, it instead maps to: 92.168.99.100:3000
which I only found out by running: `docker-machine env default` found here: https://github.com/docker/kitematic/issues/1010
is my docker setup messed up?

@toolness maybe get someone else to verify localhost:3000 works for them.  Other than that looks good 👍 So as soon as someone else can get localhost:3000 to work (assuming my docker is messed up), I can merge it in.

PR #1520, opened 2017-08-05
fix URL references to refer to openopps repo

ultrasaurus:
main repo used to be at 18F, now that one is a fork of this one
so URLs should refer to https://github.com/openopps


rogeruiz:
will merge once it passes.

PR #1524, opened 2017-08-19
Re-enabling the localize call so those values can be changed via the theme.

DavidEBest:
Just uncommenting a few lines of code.


ultrasaurus:
looks good!

PR #1538, opened 2017-09-07
Should fix issue #1345

### LINKED ISSUE #1345: an Opportunity requestor should be able to remove a participant
this works for an administrator, but as an Opportunity creator I get prompted to login when I remove a participant and save

bmcorum:
Property name on submitted mode is task not taskId. This fix should also fix one area causing application crashes. Resolves #1345.


lawrencek76:
I have tested locally by logging in as a user and accepting an opportunity and then logging in as the opportunity owner and clicking the x by the first users name and no errors were generated.

PR #1547, opened 2017-09-14
Add platform specific support to package scripts.

bmcorum:
Still need to finish scripts for Windows platform and test.


lawrencek76:
This approach appears to double the burden of maintaining the scripts for everyone.  At this point the *nix versions have to work so it seems while this does make it work on windows it does so at the cost of maintainability meaning that it will be easy to build scripts that result in different and unexpected behavior from one os to the other, particularly because who ever modifies a script for one os is very likely not familiar with the other.   (The opposite of falling into the pit of success)

A preferable method would be some sort of interpreter that parses the original command and translates it to a compatible windows command and generates an error on *nix platforms if it does not know how to do so ensuring a higher level of consistency and some test ability across platforms.

Also there is WSL on the windows side to consider.

bmcorum:
@lawrencek76 time is not on our side at this point and ignoring a potential solution just because it "seems" like a burden as opposed to not working at all would be counter productive. We need a working solution and once we have that and can stabilize the solution, then we can look at better solutions.

bmcorum:
researching updating the npm-scripts to call out to a NodeJS file so that we are not utilizing any operating specific commands or tools.

lawrencek76:
What I want here is a discussion about alternatives and why we choose one over another.  This solution has very real consequences (almost all do) and it would be nice to know why we choose to accept one solution over others and acknowledged the benefits and new challenges we are taking on when we know them advance. 

So in this case in I would like to know did we look at the other 2 ideas and why they do or don't work for us. And then if we do go this way is there anything we can do to mitigate the risk or maybe the risk is minor and we just accept that it may happen.

The goal of this discussion would be that in the future if someone is looking into why or how we are doing something they could look at what was considered and realize that either a new option C now exists or we already looked at option C and here is why it did not work for us at the time.

Ignoring is completely the opposite of the idea.

bmcorum:
So far I have a install-theme.js that uses node to clone a defined theme repository and then copy the required files. This will replace the preinstall and import npm scripts and should run regardless of the host system.

lawrencek76:
Closing this will use approach from pr #1549

PR #1549, opened 2017-09-14
agnostic install

bmcorum:
preinstall, import, postinstall and uswds no longer uses platform command line, but instead uses NodeJS for installing THEME and copying uswds files.


bmcorum:
@lawrencek76 in order to revert the changes to the .cfignore file we would need to update the THEME package to change the file type of the .cfignore file in the THEME project to be a file not a link.

bmcorum:
@lawrencek76 the cfignore file has been reverted and the package JSON has been updated to remove the unnecessary call to node version, current working directory, and directory listing upon completion of an npm install command.

.cfignore
```diff
@@ -24,7 +24,6 @@ config/local.js
 # Ignore production files
 assets/build
 assets/uploads
-#assets/js/bundle.js
```
rogeruiz: 😅

bmcorum:
@rogeruiz, yes this should allow for the project to be installed and run on any host OS. I developed the branch on a Mac to make sure I did not break anything related to a UNIX based system and then did a fresh pull and install on a Windows machine.

PR #1588, opened 2017-12-21
change PUTS to POST in koa branch

### LINKED ISSUE #1589: OPM & Akamai WAF
In order to on-board the openopps-platform at OPM, the application must be behind Akamai. The default configuration for WAF is to block all HTTP PUT and DELETE request. Need to update the openopps-platform to not use HTTP PUT and DELETE verbs. Recommend updating to use POST verb, and additional values in the request body to provide the same functional outcome of the equivalent PUT or DELETE.

Marchalk:
Fixes #1589 to update PUT to a POST since PUT requests are blocked by Akamai WAF.

### DIFF
--- a/api/opportunity/controller.js
+++ b/api/opportunity/controller.js
@@ -31,10 +31,10 @@ router.get('/api/task/:id', async (ctx, next) => {
   if (typeof ctx.state.user !== 'undefined' && ctx.state.user.id === task.userId) { 
     task.isOwner = true; 
   }
-  if (task.isOwner || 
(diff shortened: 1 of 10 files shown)


lawrencek76:
This cannot be merged as is.  It has failed to meet many of the [contribution policies](https://github.com/openopps/openopps-platform/blob/dev/CONTRIBUTING.md).  
Most notable is no descriptive commit message and no issue being fixed.  We need an issue that explain the motivation as the guidelines also say.

PR #1592, opened 2017-12-27
Update mac install with dbmigrate command..

ultrasaurus:
I tested install instructions starting from no database, and realized we were missing the db migrate command.

Also, added some instructions that we need to "ctrl-c" after the init and demo scripts, since I thought that might be helpful for new folks.


bmcorum:
@ultrasaurus instead of having the user do a ctrl+c, should we just call something like process.exit() so we have a more graceful exit?

PR #1609, opened 2018-01-03
Set up cache control configuration.

bmcorum:
Added new file config\cache.js. Currently utilizing cache.public and
cache.noStore options.


lawrencek76:
looks like a good place to start

PR #1632, opened 2018-01-07
refactor logging code

ultrasaurus:
- make it so we can `use('log')` so logging is easier to modify
  - added npm module [@iarna/lib](https://www.npmjs.com/package/@iarna/lib) which allows using the same line of code to include logging anywhere in the directory hierarchy, avoiding `require('../../..`
  - in the future, we can easily modify logging code in one place, if/when we want to switch logging modules or add writing to a file or whatever else we might need
- refactored code so that files `use('log')` to load shared logging code in  `lib/log` instead of blue-ox directly


ultrasaurus:
oops -- looks like the library I want to use is missing a license file.  I ping'd the author, hopefully that will get updated so we can use it.

ultrasaurus:
Note: Version Eye was incorrect, there was a license all along!  

https://twitter.com/VersionEye/status/954383974343806977

bmcorum:
@ultrasaurus sad day that versioneye is shutting down. Guess I will research an alternative.

ultrasaurus:
rebased on top of dev -- did one manual merge in app-koa.js, so please review lib/log/middleware.js carefully to make sure the merge was correct.  I did test manually after the merge (startup and user sign up) and the logs appeared to be unchanged.

ultrasaurus:
Sorry not to see the process note earlier!  Indeed, it's important that we all communicate what we're doing, especially since we don't often overlap in time and space :)

For context, what motivated this was setting up dev server (https://github.com/openopps/openopps-platform/issues/1560 ) where I couldn't find the logs at first.  I've since learned more about docker and found a few different ways to get logs. I've written up a lower priority bug https://github.com/openopps/openopps-platform/issues/1738 to make it so we can have different logging for different environments -- this refactor was a pre-requisite in my mind (which I should have said in original PR notes!)

cc @bmcorum

PR #1663, opened 2018-01-25
Fix incompatibility with iarna/lib.

bmcorum:
Removed iarna/lib dependency and replaced with app-module-path.


ultrasaurus:
The solution you came up with looks straightforward.  What was the issues you ran into with `iarna/lib`?

ultrasaurus:
oh!  just saw that @bmcorum mentioned via slack: iarna breaks the integration test because something in the iarna library throws a null exception

Issue #1800, opened 2018-03-01
As an open opportunities user, I would like the landing page text to match the screen shot.


kbwagner:
1) On the landing page, the text on the page doesn't match the new screen shot (ticket 1680)

Current Text:
- Browse opportunities on the site to find the one that is right for you.
- Then just click "Participate" to get started.
- Post a discussion topic if you have questions.

New Text: 
- Browse opportunities on the site to find the one that is right for you.
- Then just click "Apply" to get started.
- Post a discussion topic if you have questions.

2) Also, please revisit the mock to ensure the button is moved to the right place: https://projects.invisionapp.com/d/main#/console/12657410/268909834/preview

3) Change both red buttons to say "Find opportunities" so they are consistent.

Kibrahacha67:
3-13-18 LHammond in Testing
Completed Testing

margaretdumas:
Mdumas - I accidently moved this one to ready while dragging over another ticket.

kbwagner:
3/18/18 - Katrina tested in Chrome and mobile Safari.  Looks good.
@LisaLNelson - ready for you to test.

LisaLNelson:
Good to go

LisaLNelson:
Approved for closure. LN

LisaLNelson:
Approved for closure. LN

PR #1932, opened 2018-03-12
Fix landing page text to match the screen shot

jhanson12:
Fix landing page text to match the screen shot. Fix list and button alignment for all break points.

#1800


raywilliams:
I already made some of these changes to the landing page. We need to work together to see which changes to keep.

Issue #1896, opened 2018-03-08
Bug - Mobile: Profile display issues


kbwagner:
The following display issues occur on the mobile version of the profile: 
1) Profile Page not lining up;
![image](https://user-images.githubusercontent.com/18709918/37167640-77495f26-22d0-11e8-9e88-06c48a79cfce.png)

2) Full title of the opportunity doesn't display in profile dashboard.
![image](https://user-images.githubusercontent.com/18709918/37177505-a2def200-22ed-11e8-860e-0097e0ae0f4a.png)

bmcorum:
@margaretdumas and @Kibrahacha67 this is ready for testing

Kibrahacha67:
3-15-18 LHammond In Testing

margaretdumas:
Mdumas - In testing

margaretdumas:
Sorry, ticket is already in testing by Larry.

kbwagner:
3/19/18 - Katrina Reviewed on mobile Safari.  Looks good.
@LisaLNelson - ready for your review.

LisaLNelson:
Tested. Not displaying properly on my Samsung Galaxy. Sending back to In Development.

LisaLNelson:
Approved for closure LN

Issue #1883, opened 2018-03-07
Bug: Mobile - Unable to click on opportunity created from profile page


kbwagner:
I am unable to click on the opportunity I created from my profile page

bmcorum:
@margaretdumas and @Kibrahacha67 this is ready for testing

Kibrahacha67:
3-15-18 LHammond In Testing

LisaLNelson:
Tested and not working on my Samsung Galaxy. Returned to In Development.

margaretdumas:
Mdumas - Cannot reproduce.  Please clear cache.

KatMullens-work:
I was able to click on a Submitted Opportunity and an Open Opportunity from my profile page. 

I created the opportunity from the button on the /tasks page, not from the profile page. I don't see an "add opportunity", etc. button on the profile page, though, so I think this passed the test! :)

LisaLNelson:
Approved for closure LN

PR #1949, opened 2018-03-15
Update Profile Edit Page For Mobile

raywilliams:
- Update Profile edit page to correct mobile display issues

#1883 
#1896


margaretdumas:
added notes to wrong ticket.

Issue #1934, opened 2018-03-13
Help Center location - landing page and footer


JodieYoung1:
As an Open Opportunities user, I want to be able to click on the Help Center link under "How does it work?" on the landing page as well as on the footer that reads Visit our Help Center

![Screen Shot 2018-03-12 at 10.59.28 AM.png](https://images.zenhubusercontent.com/59fb63728a75884b9085d569/dbe0f4de-ed48-40a8-b1df-b91eb1591ae0)

margaretdumas:
03/15/2018 - Mdumas - In testing....

margaretdumas:
Mdumas 03/16/2018 - Footer link different than mock.  Moving back to development.

margaretdumas:
Mdumas - Testing complete.  Needed to refresh the page.

JodieYoung1:
Tested successfully. Was able to click on the Help Center link from both the Landing Page and the Footer.

kbwagner:
Reopening since this was accidentally closed.

LisaLNelson:
Approved for closure LN

PR #1951, opened 2018-03-15
Add Help Center location - landing page and footer

jhanson12:
Add link and verbiage for help center on the home page and footer.  Add style for all break points including mobile.

#1934


LisaLNelson:
Note: No one is going to find this at the bottom of the landing page.

Issue #1697, opened 2018-02-07
Open Opps System Email <Discussion - new comment>


JodieYoung1:
As a task creator, the desired text, including personalizations/links is emailed to the the TASK CREATOR when the Trigger happens.

Current Message Name: Discussion –new comment
Desired Message Name: Discussion –new comment
Message Type: Email
Header Type/Subject Copy: There's a new comment on "[opportunity name]" 
Desired Text Copy:
 Hi [[GivenName]], 

There's a new comment in your opportunity.  [Name] said:

"comment"

[Take a moment to reply](link/button).

Thanks,
The Open Opportunities Team

Trigger: When a user posts new comment

Acceptance Criteria:
The desired text, including personalizations/links is emailed to the the TASK CREATOR when the Trigger happens.

Trigger: When a user posts new comment

Desired Text Copy:
 Hi [[GivenName]], 
There's a new comment in your opportunity.  [Name] said:

"comment"

[Take a moment to reply](link/button).

Thanks,
The Open Opportunities Team

bmcorum:
@margaretdumas and @Kibrahacha67 this is ready for testing

Kibrahacha67:
4/3/18 LHammond in testing

LisaLNelson:
Moving to ready to deploy LN

LisaLNelson:
Approval for closure. LN

PR #2008, opened 2018-03-30
Update New Comment Email

raywilliams:
- Update email sent to task owner when there is a new comment.

#1697


JodieYoung1:
1697 was tested successfully.

Issue #2830, opened 2019-01-10
BUG: Opportunity restrict to agency


bmcorum:
When editing an existing opportunity the restrict to agency employees checkbox does not display.

margaretdumas:
1/10/19 - Mdumas - Passed testing. Moving on to review.

kbwagner:
1/15/19 - KWagner confirmed in Test/Chrome that if I edit an existing opportunity the restrict to agency checkbox does display
- Ready for final review.

PR #2831, opened 2019-01-10
BUG: Opportunity restrict to agency

bmcorum:
Fix issue where the restrict to agency employees checkbox was not being
displayed when editing an existing opportunity.

#2830


bmcorum:
@margaretdumas and @Kibrahacha67 this is ready to test.

Issue #2739, opened 2018-12-10
Admin: when changing a owner, modal does not close and page does not update...even though the change took place


Kibrahacha67:
Steps to reproduce, In Test, follow these steps
--sign in as a Administrator
--click Administration link
--click agency
--select OPM as Agency to utilize
--click on Manage Opportunities
--click Open radio button 
--select a particular opportunity to change owner
--click the change owner icon on the right hand side of the screen
--search for new owner via drop down
--click the change owner

Kibrahacha67:
6-28-19 LHammond in testing.

kbwagner:
7/9/19 - KWagner tested in Test/Chrome
Ready for final review.

PR #3662, opened 2019-06-26
Admin: when changing a owner, modal does not close and page does not …

tanichacko:
…update...even though the change took place

- Change queries from  tagentity table to agency table

#2739


Kibrahacha67:
6-28-19 LHammond in testing

